PR #17, opened 2023-01-20
Bump nixpkgs to nixos-22.11

danielfullmer:
###### Description of changes

Bumped nixpkgs to NixOS 22.11

###### Testing

Tested an Orin AGX system can switch from 22.05 to 22.11. Need to test flashing and installer ISO.


danielfullmer:
Tested flashing and installer ISO works as well.

Issue #16, opened 2023-01-17
eMMC Flash Support


sphw:
I'm interested in trying to add support for generating and flashing an eMMC image to `jetpack-nix`. It seems like it should not be a huge undertaking. Are you guys planning on adding support? And/or would you be welcome to a contribution?

There are a few subtleties that I'm interested in before I go off and start developing. It seems like `./flash.sh` expects a separate `esp.img` and `system.img` when writing to eMMC, but Nix's built in image builder creates a combined image. I suspect that you can just use the built-in L4T ESP image, but I'm not sure if that will break things. Any help or guidance you all could offer would be great.

danielfullmer:
So, my goal with this was to have the flashing script just create a system that's ready to boot via UEFI like any other x86/arm UEFI system.  Upstream jetpack actually has additional partitions that it flashes to the QSPI/emmc which I remove (kernel/kernel_b/APP/etc): https://github.com/anduril/jetpack-nixos/blob/a6079bc8edce859ea32cdfeb7498af9b2d0a6395/modules/devices.nix#L39-L45
Their `esp.img` includes an EFI application called L4TLauncher.efi that enables backwards compatibility with the way booting worked on pre-Jetpack 5 devices, e.g. using an extlinux.conf file.  I personally didn't see any need for this and just wanted to stick with standard UEFI booting on NixOS (systemd-boot/grub/etc).  Actually flashing the device with a NixOS system was outside the scope of what I was intending.

While I wasn't planning on adding support for this, but I'd be curious to hear why you'd like this. Is it just to flash everything using one command? We currently have a two-step process for flashing where we first run this flashing script and then boot via usb or the network to partition and install a NixOS system.

Additionally, I believe the Orin NX only has QSPI, no eMMC at all. So I don't believe it'd be possible to flash a full NixOS system using the flashing script.

sphw:
> While I wasn't planning on adding support for this, but I'd be curious to hear why you'd like this. Is it just to flash everything using one command? We currently have a two-step process for flashing where we first run this flashing script and then boot via usb or the network to partition and install a NixOS system.

It is mainly to support a single step installation. I'd also like to be able to use the Jetson OTA update system to update the entire image at once. Would you still be open to this feature being added, even though it's not your exact target use case?

> Additionally, I believe the Orin NX only has QSPI, no eMMC at all. So I don't believe it'd be possible to flash a full NixOS system using the flashing script
[...]

PR #18, opened 2023-01-20
Fix NixOS booting issue on Jetson AGX Orin with Jetson Linux 34

nesteroff:
###### Description of changes

Currently some Jetson AGX Orin development kits are shipped with Jetson Linux 34. After UEFI firmware is updated using flash-orin-agx-devkit script the system fails to boot from a USB stick with NixOS ISO image.

As it turned out the reason why NixOS fails to boot is that UEFI firmware [loads device tree from the EMMC](https://github.com/NVIDIA/edk2-nvidia/blob/main/Silicon/NVIDIA/Library/DxeDtPlatformDtbLoaderLib/DxeDtPlatformDtbKernelLoaderLib.c#L525) even though we are booting from a USB device. The flash script only updates QSPI because EMMC entry is completely deleted from the flash.xml using this XPath expression: `'//device[@type="sdmmc_user"]’`. It means that in this case we still have device tree from version 34 on EMMC but NixOS is built based on version 35 and it fails to boot because the device tree is not compatible.

The simplest fix would be to keep kernel-dtb partitions in the flash.xml so that the flash script updates device tree on EMMC as well. Another idea is to change the XPath expression to `'//device[@type="sdmmc_user"]/*’` so that it keeps an empty sdmmc_user tag to make the flash script erase the EMMC. When EMMC is erased the UEFI firmware will use the device tree from the QSPI. However the problem with these ideas is that they both make changes to the EMMC which may not be expected when flashing a bootloader.

In order to not to make any changes to EMMC I decided to create a small patch for UEFI firmware so that it doesn't try to load the device tree from EMMC at all. With this patch it will use the device tree from QSPI.

We could also try to copy device tree files to the ISO image and then load it in GRUB but it's probably a lot harder to implement especially considering that the same images is supposed to work on different models.

###### Testing

- Flash [Jetson Linux 34.1](https://developer.nvidia.com/embedded/jetson-linux-r341) on Jetson AGX Orin devkit.
- Flash UEFI firmware.
- Insert a USB stick with NixOS ISO image.
- NixOS fails to boot but with this patch it boots fine.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -114,7 +114,7 @@ in rec {
     hardware.nvidia-jetpack.enable = true;
     networking.hostName = n; # Just so it sets the flash binary name.
   }).config)) {
-    "orin-agx-devkit" = { som = "orin-agx"; carrierBoard = "devkit"; };
+    "orin-agx-devkit" = { som = "orin-agx"; carrierBoard = "devkit"; bootloader.edk2NvidiaPatches = [ ./edk2-uefi-dtb.patch ]; };
     "xavier-agx-devkit" = { som = "xavier-agx"; carrierBoard = "devkit"; };
     "xavier-nx-devkit" = { som = "xavier-nx"; carrierBoard = "devkit"; };
     "xavier-nx-devkit-emmc" = { som = "xavier-nx-emmc"; carrierBoard = "devkit"; };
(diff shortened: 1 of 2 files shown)


danielfullmer:
This is an excellent writeup of the issue and solution! Thanks!

I agree with the approach taken here--given we're just trying to turn these devices into standard UEFI-based systems,  it makes most sense to use the device tree included with UEFI instead of using a kernel-dtb partition from the eMMC.  If we later want to make this like a more standard L4T setup (like in https://github.com/anduril/jetpack-nixos/issues/16), we might make this conditional.

danielfullmer (approved):

PR #19, opened 2023-01-24
Add realtime.kernel option to enable PREEMPT_RT patches

danielfullmer:
###### Description of changes

NVIDIA seems to have some level of support for running a kernel with the PREEMPT_RT patchset. They include the patches under `rt-patches/` and have instructions to enable the realtime kernel here:
https://docs.nvidia.com/jetson/archives/r35.1/DeveloperGuide/text/SD/Kernel/KernelCustomization.html#to-build-the-real-time-kernel

This adds a `hardware.nvidia-jetpack.kernel.realtime` option which, if set to true, enables the PREEMPT_RT patches and kernel options.

###### Testing

Only tested this builds. Will test on device soon.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -55,10 +55,13 @@ let
   samples = callPackages ./samples.nix { inherit debs cudaVersion autoAddOpenGLRunpathHook l4t cudaPackages; };
 
   kernel = callPackage ./kernel { inherit (l4t) l4t-xusb-firmware; };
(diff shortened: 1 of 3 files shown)


danielfullmer:
Tested working on a Xavier AGX. Some basic tests using `cycletest` show decreases in maximum latency from ~10ms down to ~50μs.

danielfullmer:
Going to merge this. Anyone who wants to use this should just be aware that (for the forseeable future) it's going to be a configuration that is less well tested than the standard configuration.

PR #21, opened 2023-02-02
Fix NVRTC and build NVRTC samples

danielfullmer:
###### Description of changes

The NVRTC (Runtime Compiler) in `libnvrtc.so` wasn't correctly loading `libnvrtc-builtins.so` at runtime, since its own lib directory was missing from the RPATH.  Additionally, this PR fixes building the `*_nvrtc` examples from cuda-samples.

###### Testing

Tested by running the following on a Xavier AGX:
```
$ nix build .#samples.cuda-samples
$ cd ./result/data
$ ../bin/clock_nvrtc
CUDA Clock sample
> Using CUDA Device [0]: Xavier
> Using CUDA Device [0]: Xavier
> GPU Device has SM 7.2 compute capability
Average clocks/block = 3272.921875
```


danielfullmer:
Also tested working on an Orin AGX.

PR #23, opened 2023-02-09
Add python-jetson package

keegandahm:
###### Description of changes

This adds a package definition for [python-jetson](https://github.com/NVIDIA/python-jetson)

###### Testing

This package has been tested against a Jetson AGX Xavier devkit


default.nix
```diff
@@ -52,6 +52,8 @@ let
     inherit bspSrc l4tVersion;
   };
 
+  python-jetson = callPackage ./python-jetson.nix { };
```
danielfullmer, line 55: Could you switch this `callPackage` to `python3.callPackage`? Then your default.nix shouldn't need to accept `python3` as an argument, but rather `buildPythonPackage` and `pyftdi` directly.

danielfullmer (changes requested):

default.nix
keegandahm: Good improvement, I just updated my PR

danielfullmer:
Tested working on my Xavier AGX.  Added a couple small improvements before merging.

PR #26, opened 2023-02-17
Add the ability to flash NixOS to eMMC on AGX Orin

### LINKED ISSUE #16: eMMC Flash Support

nesteroff:
###### Description of changes

This pull request adds the ability to flash NixOS to the eMMC on Nvidia AGX Orin.

In this implementation I'm copying boot and rootfs partitions from the sd-image to the place where the flashing script expects them to be and passing the -r flag so that it doesn't try to rebuild them. I decided to use L4TLauncher as a UEFI bootloader in order to make it similar to how it is done in Nvidia SDK and for simplicity. However it shouldn't be to hard to switch to systemd-boot or something similar if needed.

I'm not sure if this is the best possible implementation because I'm still new to Nix. Feel free to improve it or suggest changes.

Hopefully it resolves #16 

###### Testing

Tested on Nvidia AGX Orin Developer Kit using the following commands:
```
nix build .#flash-scripts.flash-orin-agx-devkit-emmc
sudo ./result/bin/flash-orin-agx-devkit
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -89,7 +89,7 @@ in rec {
   # GST plugins
 
   # Generate a flash script using the built configuration options set in a NixOS configuration
-  flashScriptFromNixos = config: let
+  flashScriptFromNixos = flashEmmc: config: let
     cfg = config.hardware.nvidia-jetpack;
   in callPackage ./flash-script.nix {
     name = config.networking.hostName;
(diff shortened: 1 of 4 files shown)


nesteroff:
Closing this one since it's already outdated and can't be merge. It can be fixed and improved later on in case there is any interest in this feature in the community.

PR #27, opened 2023-02-18
cudaPackages: fix discovery of gcc libraries for cross builds

nagisa:
###### Description of changes

In cross-compile scenarios, the directory containing `libgcc_s.so` and other such
libraries is actually under a target-specific directory such as
`${stdenv.cc.cc.lib}/aarch64-unknown-linux-gnu/lib/` rather than just plain `/lib` which
makes `autoPatchelfHook` fail at finding them libraries.

###### Testing

I cross-built a system that contains `tensorrt` in its `systemPackages`, deployed it, verified that `trtexec` manages to print out its usage when invoked without arguments.


danielfullmer:
Thanks! I added a small modification so that this only affects cross-builds and doesn't change the hash for native builds.

Is there an upstream issue for this in nixpkgs? If not, would you create one? It'd be nice to have this fixed upstream.

PR #28, opened 2023-02-20
Enable IPv6 NF_MATCH kernel parameters

nagisa:
###### Description of changes

This ensures that systems with IPv6 configured can successfully start their `firewall.service`s (at least on master nixpkgs).

###### Testing

I built a system with `networking.enableIPv6` and checked `systemctl status firewall` which is now successful.


keegandahm:
I was able to successfully test this with an Orin AGX, and the firewall service seems to work correctly with this kernel.

danielfullmer (approved):

PR #32, opened 2023-03-03
Support for Orin NX

danielfullmer:
###### Description of changes

This PR adds support for the Orin NX, running on the Xavier NX devkit.

There are two known issues:
1) Flashing a factory Orin NX works the first time, but fails the second time: https://github.com/anduril/jetpack-nixos/issues/33
2) It currently requires the USB_OTG kernel config option to be disabled, or else the kernel crashes at boot: https://github.com/anduril/jetpack-nixos/issues/34

###### Testing

Flashed and booted NixOS on an Orin NX

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,12 @@ This repository packages components from NVIDIA's [Jetpack SDK](https://develope
 
 This package is based on the Jetpack 5 release, and will only work with devices supported by Jetpack 5.1:
  * Jetson Orin AGX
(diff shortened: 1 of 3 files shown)


danielfullmer:
Merging this PR even though there are two serious known issues. We'll resolve them in future PRs when we figure out a solution.  Orin NX in jetpack-nixos should be considered _experimental_ until we have these fixed.

PR #37, opened 2023-03-08
Added initial jetson benchmarks

ktrinh-anduril:
###### Description of changes
Added support for building the jetson benchmark result from nvidia at: https://github.com/NVIDIA-AI-IOT/jetson_benchmarks/, these benchmarks are valuable since Nvidia publishes expected results that allow users to compare whether their platform is performing as expected

###### Testing
Tested on Orin NX. The benchmark is built differently based on whether it runs on NX, Orin AGX, Xavier, AGX, etc. Right now only the one for NX is predefined but others can be defined as well

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -62,6 +62,19 @@ let
 
   rtkernel = callPackage ./kernel { inherit (l4t) l4t-xusb-firmware; realtime = true; };
   rtkernelPackages = (pkgs.linuxPackagesFor rtkernel).extend kernelPackagesOverlay;
+
+  nxJetsonBenchmarks = pkgs.callPackage ./jetson-benchmarks/default.nix {
+    targetSom = "nx";
+    inherit cudaPackages;
+  };
+  xavierAgxJetsonBenchmarks = pkgs.callPackage ./jetson-benchmarks/default.nix {
+    targetSom = "xavier-agx";
+    inherit cudaPackages;
+  };
+  orinAgxJetsonBenchmarks = pkgs.callPackage ./jetson-benchmarks/default.nix {
+    targetSom = "orin-agx";
+    inherit cudaPackages;
+  };
 in rec {
   inherit jetpackVersion l4tVersion cudaVersion;
 
@@ -75,6 +88,8 @@ in rec {
 
   inherit kernel kernelPackages;
   inherit rtkernel rtkernelPackages;
+
+  inherit nxJetsonBenchmarks xavierAgxJetsonBenchmarks orinAgxJetsonBenchmarks;
 
   # TODO: Source packages. source_sync.sh from bspSrc
   # OPTEE
--- a/jetson-benchmarks/default.nix
+++ b/jetson-benchmarks/default.nix
@@ -1,0 +1,114 @@
+{ stdenvNoCC
+, fetchFromGitHub
+, python3
+, runCommand
+, lib
+, writeShellApplication
+, writeShellScript
+, coreutils
+, unzip
+, wget
+, gnupatch
+, cudaPackages
+, makeWrapper
+
+, targetSom
+
+, benchmarkSrc ? fetchFromGitHub {
+    owner = "NVIDIA-AI-IOT";
+    repo = "jetson_benchmarks";
+    rev = "43892b9ec64abdfabb4c18e19f301d9d4358f5ea";
+    sha256 = "sha256-u11iBbEALOMite/ivm95TAnmXB71i9OjdNnRd0e1cHg=";
+  }
+
+  # disable some system checks such as closing all apps which prevent the benchmark
+  # from being run non-interactively and others like the ones that set the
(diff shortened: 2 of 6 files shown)


jetson-benchmarks/scripts/run-all-jetson-benchmarks
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+script_dir=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+${script_dir}/run-jetson-benchmarks --all
```
danielfullmer, line 5: Your editor is not adding trailing newlines to these files.

jetson-benchmarks/default.nix
```diff
@@ -0,0 +1,81 @@
...
+
+  patches = [ ]
+    ++ lib.optional disableDisruptiveSystemCheck ./0001-disable-disruptive-system_check.patch;
+
+  nativeBuildInputs = [ coreutils makeWrapper ];
+
+  dontBuild = true;
+
+  installPhase = ''
+    substituteInPlace utils/load_store_engine.py --replace "/usr/src/tensorrt" "${cudaPackages.tensorrt}"
```
danielfullmer, line 64: These substituteInPlaces should be in `postPatch`

jetson-benchmarks/default.nix
```diff
@@ -0,0 +1,81 @@
...
+    cp -r $models/. models
+
+    mkdir -p $out
+    cp -r ./. $out
+
+    mkdir -p $out/bin
+    cp -r ${./scripts}/. $out/bin
+    chmod +x $out/bin/*
+
+    wrapProgram $out/bin/run-jetson-benchmarks \
```
danielfullmer, line 77: I'd put this `wrapProgram` in `postFixup`

default.nix
```diff
@@ -68,6 +68,11 @@ let
 
   rtkernel = callPackage ./kernel { inherit (l4t) l4t-xusb-firmware; realtime = true; };
   rtkernelPackages = (pkgs.linuxPackagesFor rtkernel).extend kernelPackagesOverlay;
+
+  nx-jetson-benchmarks = callPackage ./jetson-benchmarks/default.nix {
```
danielfullmer, line 72: Could you go ahead and make the other variants?

jetson-benchmarks/scripts/run-jetson-benchmarks
```diff
@@ -0,0 +1,17 @@
...
+
+script_dir=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+WORKDIR=$(mktemp -d)
+function on_exit() {
+    sudo rm -rf "$WORKDIR"
+}
+trap on_exit EXIT
+
+cp -r ${script_dir}/../. "$WORKDIR"
```
danielfullmer, line 13: Could you explain what's going on here?

danielfullmer (changes requested):

danielfullmer:
Also, could you modify the commit title to start with a capital letter?

jetson-benchmarks/scripts/run-jetson-benchmarks
ktrinh-anduril: This script will be located at ${out}/bin, when invoked, it will copy everything in the $out dir to a temp folder and start the benchmark from there. We have to do this since the benchmark generates file so can't be run from a read-only context.

danielfullmer: Makes sense. Thanks!

default.nix
ktrinh-anduril: Made other variants and tested them on the hitl rack

danielfullmer:
Could you squash these commits?

modules/jetson-benchmarks.nix
```diff
@@ -0,0 +1,30 @@
...
+let
+  cfg = config.jetson-benchmarks;
+
+  benchmarkPackage = pkgs.callPackage ../jetson-benchmarks/default.nix {
+    targetSom = config.hardware.nvidia-jetpack.som;
+    cudaPackages = pkgs.nvidia-jetpack.cudaPackages;
+  };
+in
+{
+  options.jetson-benchmarks = {
```
danielfullmer, line 16: This module isn't necessary. It doesn't modify the system other than putting the benchmark package into environment.systemPackages and that can be done by the user if they want

danielfullmer (changes requested):

modules/jetson-benchmarks.nix
ktrinh-anduril: Sorry about that, forgot to remove it

danielfullmer (approved):

Issue #38, opened 2023-03-09
Hash mismatch building nvidia-display-driver


mikatammi:
The nvidia-display-driver fetchgit uses a tag instead of a commit hash. Now NVIDIA changed the tag to point to a different place, leading to a different content

How to replicate:
```
$ nix build .#packages.aarch64-linux.iso_minimal
warning: Git tree '/home/mika/anduriltest/jetpack-nixos' is dirty
error: hash mismatch in fixed-output derivation '/nix/store/rr2cyni6sj6gggs3hrzdkwh88ar906ng-nv-kernel-display-driver-5f54f1d.drv':
         specified: sha256-fP54ztq4oNKgfn6LckqsOuPjYE+SB1W083DOwRNth80=
            got:    sha256-ZY3eaP6dko6GnagcJIwAEb+aZ+IYxA+SsiaYBGK42oA=
error: 1 dependencies of derivation '/nix/store/msnrzjfsadl5pfxaphbwzk6yxxbpw47w-nvidia-display-driver-jetson_35.2.1.drv' failed to build
error: 1 dependencies of derivation '/nix/store/0jpjj04z39bvh67p3avj37lrqrvyzsyf-kernel-modules.drv' failed to build
error: 1 dependencies of derivation '/nix/store/v6b9mgw3pmmhfrsxjx17rbjhkcjgg3w2-linux-5.10.104-modules.drv' failed to build
error: 1 dependencies of derivation '/nix/store/5aaz5xfsgr04jfcfrnpkbym2rg9qrcdr-linux-5.10.104-modules-shrunk.drv' failed to build
error: 1 dependencies of derivation '/nix/store/cbxmqzlswh8hkm90xnjcch3fgwl7y5cc-nixos-system-nixos-22.11pre-git.drv' failed to build
error: 1 dependencies of derivation '/nix/store/zhin4s390wdb1nnv0wfbvv76b909j3rh-closure-info.drv' failed to build
error: 1 dependencies of derivation '/nix/store/fdfl21q1jf4y37h7hkyaipz0s09r26iy-efi-directory.drv' failed to build
error: 1 dependencies of derivation '/nix/store/fx0ay5dc5ax6sdlngbmnsic49gd20pdf-stage-1-init.sh.drv' failed to build
error: 1 dependencies of derivation '/nix/store/jkpvbqdw08wvaffciq3cyj5n0mdp69w8-nixos-22.11pre-git-aarch64-linux.iso.drv' failed to build
```

[...]

PR #39, opened 2023-03-09
Fix nvidia-display-driver

mikatammi:
###### Description of changes

Why: nvidia-display-driver package broke
* NVIDIA changed their tag to point to a different commit hash, resulting in hash mismatch. Use commit hash instead of the tag to point to a revision
* Fix directory name where the source package gets unpacked

###### Testing

Tested using AGX Orin Devkit. I was able to launch Weston, and graphics were usable in kmsdrm mode. I didn't test the legacy mode which supports X11.


mikatammi:
Fixes #38

unbel13ver (approved):

Issue #59, opened 2023-04-08
Display driver not building on latest versions of nixos-unstable


mikatammi:
It is hard to bisect because there are other erros as well, but with nixos-unstable, nvidia-display-driver builds start to fail somewhere after this commit https://github.com/NixOS/nixpkgs/commit/21ec906463ea8f11abf3f9091ddd4c3276516e58 where they change GCC 9 -> 12

mikatammi:
```
https://cache.nixos.org/d9yfr4krv9a5zb64gh59zvggwlff002h.narinfo
https://cache.nixos.org/qsa6mrzxknzii6dj44a9br53zx256msq.narinfo
https://cache.nixos.org/8389z1h34dhj4j2vx78dk9qr34vp3bml.narinfo
https://cache.nixos.org/v1lr69jb7z4l2wn7r26qhjp1qc1w6l8k.narinfo
https://cache.nixos.org/nar/0cwyd1xnx7vqhxd07khl6s9ch68x3y3m2bd71kfj039r13a55hzz.nar.xz
https://cache.nixos.org/nar/0n81dvcp2bhx2fdnn5g4fxm6sq4n3ayc6233zddl06lahkkvdwkn.nar.xz
https://cache.nixos.org/nar/1n6mlbyh0nzfyidjrva5fbb75zb9r5qbzahnyagc5k6075hasv3f.nar.xz
https://cache.nixos.org/nar/15hxj8dc6yna1yapq2diyx9vkpf8qha4282gjf8yv4qmibmqpvi5.nar.xz
https://cache.nixos.org/nar/02aszxqrq88g7ji4a73sjch9xj3flbmnjdq8w1ms45f60ichgxf9.nar.xz
https://cache.nixos.org/nar/015j88ys5cnxh52xj28w0gly969gdf7mgwdqbg5py4zfd035pc6y.nar.xz
https://cache.nixos.org/nar/0rg5hc3crxdhr0l6g3wy5zr4xryn4hcy0130dyv2ak8xzabwsv0c.nar.xz
https://cache.nixos.org/nar/1q48341p88gijz5iixzdn8pvxfbimpc0v2r59mvhlz1bxg3q0bs8.nar.xz
https://cache.nixos.org/nar/17s6fsnx0mn7d1p4jdalnlmp6g2ykgq55qwmsprcqv1s8x0rd4pv.nar.xz
[...]

PR #60, opened 2023-04-08
Fix display-driver build on GCC 12

mikatammi:
Recent versions of nixos-unstable have moved from GCC 9 to 12, which breaks to build. Flag -mno-outline-atomics is needed for the kernel module, to avoid errors sucs as:

ERROR: modpost: "__aarch64_cas4_sync" [nvidia.ko] undefined!
ERROR: modpost: "__aarch64_ldadd8_sync" [nvidia.ko] undefined!
ERROR: modpost: "__aarch64_ldadd4_sync" [nvidia.ko] undefined!
ERROR: modpost: "__aarch64_cas8_sync" [/nvidia.ko] undefined!

This change has been backported from jetson_35.3.1

###### Description of changes

**What has changed as a result of this PR? Why was the change made?**

display-driver did not build with nixos-unstable channel. A small change adding one flag to kernel module build was backported from 35.3.1 to 35.2.1 to make it builable with recent nixos-unstable versions.

###### Testing

Tested that it builds with latest nixos-unstable, which is an improvement because previously it didn't build at all. I don't at this very moment have physical access to the AGX Orin with display to see does the driver really work or not

### DIFF
--- a/kernel/add-mno-outline-atomics-for-aarch64.patch
+++ b/kernel/add-mno-outline-atomics-for-aarch64.patch
@@ -1,0 +1,15 @@
+diff --git a/src/nvidia/Makefile b/src/nvidia/Makefile
+index 17f680e..a38eeed 100644
+--- a/src/nvidia/Makefile
++++ b/src/nvidia/Makefile
+@@ -79,6 +79,7 @@ ifeq ($(TARGET_ARCH),aarch64)
+   CFLAGS += -mgeneral-regs-only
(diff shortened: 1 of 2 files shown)


mikatammi:
This fixes #59

danielfullmer (approved):
Thanks for the fix! I have some test automation set up that is currently only targeting the 22.11 branch of nixpkgs.  I'll see if I can update that to also build/test against nixpkgs-unstable.

PR #64, opened 2023-04-21
Add derivations for OPTEE and ATF

jmbaur:
###### Description of changes

OPTEE and ATF are built according to the SOC platform (t194 or t234), so it requires having hardware.nvidia-jetpack.som defined.

Changes include:
- Adding an ATF image with OPTEE and builtin nvluks-srv and hwkey-agent trusted applications that are flashed in the same per-device flash script.
- Adding the tee-supplicant service to each board running a TOS. This is used to broker communication between the REE and TEE (regular execution environment and trusted execution environment).

###### Testing

Tested on an Orin AGX devkit. Functionality between the nix-built TOS and stock TOS appears to be the same.

### DIFF
--- a/optee.nix
+++ b/optee.nix
@@ -1,0 +1,218 @@
+{ l4tVersion
+, bspSrc
+, buildPackages
+, lib
+, stdenv
+, fetchgit
+, pkg-config
+, libuuid
+, dtc
+}:
+
+let
+  atfSrc = fetchgit {
+    url = "https://nv-tegra.nvidia.com/r/tegra/optee-src/atf";
+    rev = "jetson_${l4tVersion}";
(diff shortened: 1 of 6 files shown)


optee.nix
```diff
@@ -0,0 +1,223 @@
...
+      meta.platforms = [ "aarch64-linux" ];
+    });
+
+  buildTOS = { platform }@args:
+    let
+      armTrustedFirmware = buildArmTrustedFirmware args;
+
+      opteeDTB = buildOpteeDTB args;
+
+      foo =
```
danielfullmer, line 188: Is this leftover?

jmbaur: Yes! Thanks, I missed this

PR #81, opened 2023-05-11
Patch efitools to write to eMMC FAT partition

jmbaur:
###### Description of changes

For the xavier agx where efivars is mounted read-only and EFI variable
modifications must happen through an eMMC FAT partition, use a patched
efitools so that writes from the `efi-updatevar` utility are written to
the correct location.

###### Testing

Manually tested `efi-updatevar` with the patch applied on an xavier-agx devkit

### DIFF
--- a/modules/devices.nix
+++ b/modules/devices.nix
@@ -118,5 +118,15 @@ in lib.mkMerge [{
     autoFormat = true;
     formatOptions = "-F 32 -n ESP";
   };
+
+  systemd.tmpfiles.rules = [ "d /opt/nvidia/esp/EFI/NVDA/Variables 0755 root root -" ];
+
+  nixpkgs.overlays = [
+    (final: prev: {
+      efitools = prev.efitools.overrideAttrs (old: {
+        patches = (old.patches or []) ++ [ ./efitools-nvidia-ro-efivars.patch ];
+      });
+    })
+  ];
 })
 ]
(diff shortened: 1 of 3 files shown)


modules/devices.nix
```diff
@@ -118,5 +118,15 @@ in lib.mkMerge [{
     autoFormat = true;
     formatOptions = "-F 32 -n ESP";
   };
+
+  systemd.tmpfiles.rules = [ "d /opt/nvidia/esp/EFI/NVDA/Variables 0755 root root -" ];
```
danielfullmer, line 122: What is the effect of this if /opt/nvidia/esp is not mounted?

modules/devices.nix
```diff
@@ -118,5 +118,15 @@ in lib.mkMerge [{
...
+
+  nixpkgs.overlays = [
```
danielfullmer, line 124: Shouldn't this only be applied if `som == "xavier-agx"`

uefi-firmware.nix
```diff
@@ -105,7 +105,7 @@ let
       src = edk2-src;
 
       depsBuildBuild = [ buildPackages.stdenv.cc ];
-      nativeBuildInputs = [ bc pythonEnv acpica-tools dtc ];
+      nativeBuildInputs = [ bc pythonEnv acpica-tools dtc util-linux ];
```
danielfullmer, line 108: This is an unrelated change, so I'd recommend doing this in another PR

danielfullmer (changes requested):

jmbaur:
Closed since patching one tool that wants to write to the efivarfs kinda opens a can of worms for patching all tools that want to do this.

PR #87, opened 2023-05-17
Add UEFI capsule authentication

jmbaur:
###### Description of changes

These changes allow for the EDK2 UEFI to be build with a public key that can verify capsule updates as well as for each subsequent capsule update to be built with a corresponding signing key so that update is accepted by the firmware. Enabling/disabling capsule authentication and configuring the keys used during builds are exposed via NixOS module options.

###### Testing

After building EDK2 with these changes and my own set of test keys, I verified on an Orin AGX that capsule updates signed with EDK2's test keys were not applied to the firmware and that capsule updates signed with my own test keys were applied.


README.md
```diff
@@ -133,6 +133,16 @@ The Capsule update status is one of the following integers:
...
 - 3 - Capsule install failed
 
+### UEFI Capsule Authentication
+
+To ensure only authenticated capsule updates are applied to the device, you can
+build the UEFI firmware and each subsequent capsule update using your own signing keys.
+An overview of the key generation can be found at [EDK2 Capsule Signing](https://github.com/tianocore/tianocore.github.io/wiki/Capsule-Based-System-Firmware-Update-Generate-Keys).
+
+To include your own signing keys in the EDK2 build and capsule update, make
+sure the option `hardware.nvidia-jetpack.firmware.uefi.capsuleAuthentication.eanble`
```
danielfullmer, line 143: s/eanble/enable/

capsule-authentication.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/Platform/NVIDIA/Jetson/Jetson.dsc b/Platform/NVIDIA/Jetson/Jetson.dsc
+index 39e2c1d..3d6911d 100644
+--- a/Platform/NVIDIA/Jetson/Jetson.dsc
++++ b/Platform/NVIDIA/Jetson/Jetson.dsc
+@@ -33,3 +33,7 @@
+ 
+ [PcdsFixedAtBuild]
+   gNVIDIATokenSpaceGuid.PcdPlatformFamilyName|L"Jetson"
++!if $(CAPSULE_ENABLE)
```
danielfullmer, line 9: I'd rename this to something like CUSTOM_CAPSULE_CERT

modules/flash-script.nix
```diff
@@ -50,6 +49,55 @@ in
...
+                used for signing capsule payloads.
+              '';
+            };
+
+            requiredSystemFeatures = lib.mkOption {
+              type = types.listOf types.str;
+              default = [ ];
+              description = lib.mdDoc ''
+                Additional `requiredSystemFeatures` to add to derivations which
+                make use of capsule authentication keys.
```
danielfullmer, line 97: s/keys/private keys/

uefi-firmware.nix
```diff
@@ -48,11 +54,14 @@ let
     sha256 = "sha256-EPtI63jYhEIo4uVTH3lUt9NC/lK5vPVacUAc5qgmz9M=";
   };
 
-  _edk2-nvidia = fetchFromGitHub {
-    owner = "NVIDIA";
-    repo = "edk2-nvidia";
-    rev = "ad78b07af65d41bb96839f0bbe67bb445e04272f"; # Latest on r35.3.1-updates as of 2023-05-01
-    sha256 = "sha256-PdrisHYkmBXGvfkNboVvKJnBqORiM8sUsGySj7n5Y5c=";
+  _edk2-nvidia = applyPatches {
```
danielfullmer, line 57: Can you apply this in the `edk2-nvidia` applyPatches below instead of making a new one?

uefi-firmware.nix
```diff
@@ -48,11 +54,14 @@ let
...
+    src = fetchFromGitHub {
+      owner = "NVIDIA";
+      repo = "edk2-nvidia";
+      rev = "ad78b07af65d41bb96839f0bbe67bb445e04272f"; # Latest on r35.3.1-updates as of 2023-05-01
+      sha256 = "sha256-PdrisHYkmBXGvfkNboVvKJnBqORiM8sUsGySj7n5Y5c=";
+    };
+    patches = [ ./capsule-authentication.patch ];
   };
   edk2-nvidia =
     if (errorLevelInfo || bootLogo != null)
```
danielfullmer, line 67: This might be a bug I introduced. I think we'd need to conditionally use the applyPatches version if edk2NvidiaPatches is nonempty.

uefi-firmware.nix
```diff
@@ -127,10 +136,19 @@ let
         chmod -R u+w BaseTools
       '';
 
+      inherit publicCertificateDer requiredSystemFeatures;
```
danielfullmer, line 139: This only requires the public key, which can be added to the nix store. So t his derivation doesn't need `requiredSystemFeatures`

danielfullmer: Also, why does `publicCertificateDer` get inherited here? It'd end up as an environment variable inside the build.

uefi-firmware.nix
```diff
@@ -142,6 +160,7 @@ let
         build -a ${targetArch} -b ${buildTarget} -t ${buildType} -p Platform/NVIDIA/Jetson/Jetson.dsc -n $NIX_BUILD_CORES \
           -D BUILDID_STRING=${l4tVersion} \
           -D BUILD_DATE_TIME="$(date --utc --iso-8601=seconds --date=@$SOURCE_DATE_EPOCH)" \
+          ${lib.optionalString (publicCertificateDer != null) "-D CAPSULE_ENABLE"} \
```
danielfullmer, line 163: Same note to change this as above.

modules/flash-script.nix
```diff
@@ -50,6 +49,55 @@ in
             type = types.listOf types.path;
             default = [];
           };
+
+          capsuleAuthentication = {
+            enable = mkEnableOption "capsule update authentication";
+
+            publicCertificateDer = mkOption {
```
danielfullmer, line 56: Recommend renaming these options to end with "File", to match what is used for the secure boot keys.

danielfullmer (changes requested):

uefi-firmware.nix
jmbaur: I had thought about this for awhile. While I agree that the public key can freely be added to the nix store, the option `publicCertificateDer` could be represented as `"/path/to/pubkey"` instead of say `./path/to/pubkey`, which means we would not be able to assume the file can be built on any machine. Adding `requiredSystemFeatures` to this derivation along with the derivations requiring the private keys felt more appropriate, because at the very least the public key can probably be safely assumed that it will exist on the same build machine that the private keys exist on?

This is also why I am setting `publicCertificateDer` as an environment variable that gets referenced in the build phase.

uefi-firmware.nix
jmbaur: Should be fixed now

uefi-firmware.nix
jmbaur: Should be fixed now

modules/flash-script.nix
jmbaur: I'll keep this to not be "private keys", pending the discussion below about the usage of `requiredSystemFeatures`

capsule-authentication.patch
jmbaur: CAPSULE_ENABLE seems to be a convention used by boards in edk2-platforms, but the naming does overlap with just enabling capsule updates to begin with, not about capsule update authentication. Renamed to `CUSTOM_CAPSULE_CERT`

uefi-firmware.nix
danielfullmer: I'd want to minimize the number of derivations that need to be built on a machine with requiredSystemFeatures, so I'd lean towards my approach.

Why use a environment variable instead of just substituting it in the nix string?

jmbaur: Sounds good. I removed requiredSystemFeatures on the derivation just using the public key. We can probably add a check to that option field that uses `lib.isStorePath` or something similar in a future PR.

PR #89, opened 2023-05-23
Add NixOS options for enrolling default secureboot keys

jmbaur:
###### Description of changes

Nvidia's flash script allows for enrolling the following UEFI protected variables:
- `dbDefault`
- `kekDefault`
- `pkDefault`

These changes take their BSP script `tools/gen_uefi_default_keys_dts.sh` to create a DTBO for use during the flashing procedure.

###### Testing

Tested on an Xavier AGX. After flashing with these changes, Nvidia's EnrollFromDefaultKeysApp UEFI application picks up the default keys and enrolls them as PK, KEK, db, etc. on first boot.

### DIFF
--- a/flash-script.nix
+++ b/flash-script.nix
@@ -16,6 +16,9 @@
 
   # Optional EKS file containing encrypted keyblob
   eksFile ? null,
+
+  # Additional DTB overlays to use during device flashing
+  additionalDtbOverlays ? [],
 }:
 ''
   set -euo pipefail
@@ -38,6 +41,8 @@
   export NO_ROOTFS=1
   export NO_RECOVERY_IMG=1
   export NO_ESP_IMG=1
+
+  export ADDITIONAL_DTB_OVERLAY=''${ADDITIONAL_DTB_OVERLAY:+$ADDITIONAL_DTB_OVERLAY,}${lib.concatStringsSep "," additionalDtbOverlays}
 
   ${lib.optionalString (partitionTemplate != null) "cp ${partitionTemplate} flash.xml"}
   ${lib.optionalString (dtbsDir != null) "cp -r ${dtbsDir}/. kernel/dtb/"}
--- a/modules/flash-script.nix
+++ b/modules/flash-script.nix
@@ -50,6 +50,28 @@ in
             default = [];
           };
 
+          secureBoot = {
+            enrollDefaultKeys = lib.mkEnableOption "enroll default UEFI keys";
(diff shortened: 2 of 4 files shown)


flash-tools.nix
```diff
@@ -26,7 +26,7 @@ let
       perl
     ];
 
-    patches = [ ./flash-tools.patch ./flash-tools-secureboot.patch ];
+    patches = [ ./flash-tools.patch ./flash-tools-secureboot.patch ./gen_uefi_default_keys.patch ];
```
danielfullmer, line 29: This file is being so extensively patched, I'd rather we not use their script at all and just create the dts file and compile it ourselves. The structure of the device tree overlay isn't so complicated we can't just do it ourselves.

flash-script.nix
```diff
@@ -39,6 +42,8 @@
   export NO_RECOVERY_IMG=1
   export NO_ESP_IMG=1
 
+  export ADDITIONAL_DTB_OVERLAY=${lib.concatStringsSep "," additionalDtbOverlays}
```
danielfullmer, line 45: This unfortunately means we can't pass ADDITIONAL_DTB_OVERLAY at runtime, can you have it append to the existing ADDITIONAL_DTB_OVERLASY if it exists?

modules/flash-script.nix
```diff
@@ -47,7 +47,29 @@ in
...
-            default = [];
+            default = [ ];
+          };
+
+          secureBoot = {
+            enrollDefaultKeys = lib.mkEnableOption "default UEFI keys";
+            defaultPkEslFile = mkOption {
+              type = lib.types.path;
+              description = lib.mdDoc ''
+                The path to the UEFI PK public key in ESL format.
```
danielfullmer, line 58: I think to be more accurate this should just say "The path to the UEFI PK EFI Signature List (ESL)"

modules/flash-script.nix
```diff
@@ -47,7 +47,29 @@ in
 
           edk2NvidiaPatches = mkOption {
             type = types.listOf types.path;
-            default = [];
+            default = [ ];
+          };
+
+          secureBoot = {
+            enrollDefaultKeys = lib.mkEnableOption "default UEFI keys";
```
danielfullmer, line 54: "enroll default UEFI keys"

danielfullmer (changes requested):

danielfullmer (approved):

PR #99, opened 2023-05-26
Remove publicCertificateDerFile and just use trustedPublicCertPemFile

danielfullmer:
###### Description of changes

The DER format of the public certificate can just be created at build-time based on the PEM-formatted trustedPublicCertPemFile. Since these are supposed to be the same certificate, it's less confusing if there's just one option for it.

Also, don't set default null 

###### Testing

Tested by performing a UEFI capsule update on the Orin nano devkit.

### DIFF
--- a/device-pkgs.nix
+++ b/device-pkgs.nix
@@ -47,7 +47,7 @@ let
       errorLevelInfo = cfg.firmware.uefi.errorLevelInfo;
       edk2NvidiaPatches = cfg.firmware.uefi.edk2NvidiaPatches;
     } // lib.optionalAttrs cfg.firmware.uefi.capsuleAuthentication.enable {
-      inherit (cfg.firmware.uefi.capsuleAuthentication) publicCertificateDerFile;
+      inherit (cfg.firmware.uefi.capsuleAuthentication) trustedPublicCertPemFile;
     });
 
     inherit socType;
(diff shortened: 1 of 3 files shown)


device-pkgs.nix
```diff
@@ -170,16 +170,16 @@ let
...
   # NOTE: providing null public certs here will use the test certs in the EDK2 repo
-  mkUefiCapsuleUpdate = lib.makeOverridable ({
-    trustedPublicCertPemFile ? null,
-    otherPublicCertPemFile ? null,
-    signerPrivateCertPemFile ? null,
-    requiredSystemFeatures ? [ ],
-  }: runCommand "uefi-${hostName}-${l4tVersion}.Cap" { nativeBuildInputs = [ python3 openssl ]; inherit requiredSystemFeatures; } ''
+  uefiCapsuleUpdate = runCommand "uefi-${hostName}-${l4tVersion}.Cap" {
+    nativeBuildInputs = [ python3 openssl ];
+    inherit (cfg.firmware.secureBoot) requiredSystemFeatures;
```
jmbaur, line 175: I believe this `requiredSystemFeatures` came from `cfg.firmware.uefi.capsuleAuthentication.requiredSystemFeatures`

danielfullmer: Fixed in new commit

PR #101, opened 2023-05-27
Expose mkFlashScript from device-pkgs

mikatammi:
###### Description of changes

Expose mkFlashScript from device-pkgs.nix. Thus one can now call devicePkgsFromNixosConfig from another package/flake, and get an attribute set where mkFlashScript is available. This makes it easier to create custom flash scripts, because not all the arguments need to be specified, but user of jetpack-nixos could just override the desired arguments, and use the defaults for others.

###### Testing

Tested from other project, using AGX Orin Devkit. This change does not really change any behavior in this project, but merely exposes one part of the API

### DIFF
--- a/device-pkgs.nix
+++ b/device-pkgs.nix
@@ -203,5 +203,6 @@ let
   };
 in {
   inherit (tosImage) nvLuksSrv hwKeyAgent;
+  inherit mkFlashScript;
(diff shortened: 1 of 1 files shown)


mikatammi:
What do you think about this?

danielfullmer:
Looks good to me! Thankis.

PR #103, opened 2023-06-03
Build tee-supplicant with a customizable TEE_CLIENT_LOAD_PATH

### LINKED ISSUE #100: Optee TAs
First off, thank you for this amazing work!! 

I am currently testing on the Orin AGX devkit, I tested this a week or 10 days ago and only getting to raise the question now:

I am not sure if this is a feature request or an - educate an ignorant user issue - I'm still new to nix in general, but here goes:

1)
It appears the default TAs are built in https://github.com/anduril/jetpack-nixos/blob/master/optee.nix#L84 and https://github.com/anduril/jetpack-nixos/blob/master/optee.nix#L190.

However, they are not exported in a package, either `optee-os` or a potential `optee-os-ta`.  If I scp the binaries to the target system from the /nix/store/.....ta, I can use them. Which leads to point 2.

2)
 Optee-client defaults to using /lib as the root search directory for the supplicant to load the TA from as defined by CFG_TEE_CLIENT_LOAD_PATH : https://nv-tegra.nvidia.com/r/gitweb?p=tegra/optee-src/nv-optee.git;a=blob;f=optee/optee_client/config.mk;h=24904af9cef9f504747549c3fc87fe06d27f14be;hb=5ac2ab218ba9116f1df4a0bb5092b1f6d810e8f7#l37

This is not very nix-y :) So I think that when building opteeClient https://github.com/anduril/jetpack-nixos/blob/master/optee.nix#L32 the CFG_TEE_CLIENT_LOAD_PATH should be set to the package that exports the TAs from point 1).

 3) 
For some reason the OpTEE developers are not happy to set a load path, but rather it is a load path `prefix` which then needs a ta dev_path to be combined with it - line 82:
 https://nv-tegra.nvidia.com/r/gitweb?p=tegra/optee-src/nv-optee.git;a=blob;f=optee/optee_client/tee-supplicant/src/teec_ta_load.c;h=2cb5e98170ed9dd999c45715c8113636f22a1c2b;hb=5ac2ab218ba9116f1df4a0bb5092b1f6d810e8f7#l82

Which means that the [supplicant](https://github.com/anduril/jetpack-nixos/blob/master/modules/default.nix#L169) needs to be launched with` -t <sub dir>`  to find the ta.  

Summary of steps I previously followed:
Build and flash firmware, and system image
scp /nix/store/....pkcs11.ta  <Orin board> /lib/ta
ssh <orin board>
install pkcs11-tool
Stop the tee-supplicant with systemctl
manually start tee-supplicant with `-t ta`
then run pkcs11-tool to test the TA is loaded and responding

So ultimately, the question is: which package should export the TAs? Then this can be plugged into the CFG_TEE_CLIENT_LOAD_PATH to build the client, and the subdirectory can be passed to the systemd unit when launching the supplicant. 

I had 2 proposals, not sure either is good. Either add an installPhase to buildOptee or add another package like buildOpteeTaDevKit or buildNvLuksSrv called optee-ta that exposes the tas.

Sorry! This turned into a thesis.

jmbaur:
###### Description of changes

These changes add a NixOS option for customizing the TEE_CLIENT_LOAD_PATH for tee-supplicant. Also added is a NixOS option for adding command-line arguments to tee-supplicant, helpful in this case for changing tee-supplicant's `--ta-dir`, the subdirectory under the load path where tee-supplicant will look for TAs. The default load path is set to "/var/lib/optee", but can trivially be changed to some static path built with nix. For example:

```nix
{ pkgs, ... }: {
  hardware.nvidia-jetpack.firmare.optee.clientLoadPath = pkgs.linkFarm "optee-load-path" [{
    name = "optee_armtz/my.ta";
    path = taDerivation;
  }];
}
``` 

where `taDerivation` is the path to the trusted application.

Fixes #100 

###### Testing

Tested on an orin-agx-devkit.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -111,10 +111,8 @@ let
 
   # Packages whose contents are paramterized by NixOS configuration
   devicePkgsFromNixosConfig = callPackage ./device-pkgs.nix {
-    inherit l4tVersion pkgsAarch64 flash-tools flashFromDevice edk2-jetson uefi-firmware buildTOS bspSrc;
+    inherit l4tVersion pkgsAarch64 flash-tools flashFromDevice edk2-jetson uefi-firmware buildTOS opteeClient bspSrc;
   };
-
-  devicePkgs = lib.mapAttrs (n: c: devicePkgsFromNixosConfig (pkgs.nixos c).config) supportedConfigurations;
 
   otaUtils = callPackage ./ota-utils {
     inherit tegra-eeprom-tool l4tVersion;
@@ -133,8 +131,6 @@ in rec {
 
   inherit kernel kernelPackages;
   inherit rtkernel rtkernelPackages;
-
-  inherit opteeClient;
 
   inherit nxJetsonBenchmarks xavierAgxJetsonBenchmarks orinAgxJetsonBenchmarks;
 
--- a/device-pkgs.nix
+++ b/device-pkgs.nix
@@ -1,5 +1,5 @@
 { lib, callPackage, runCommand, writeScript, writeShellApplication, makeInitrd, makeModulesClosure,
-  flashFromDevice, edk2-jetson, uefi-firmware, flash-tools, buildTOS,
+  flashFromDevice, edk2-jetson, uefi-firmware, flash-tools, buildTOS, opteeClient,
   python3, bspSrc, openssl, dtc,
   l4tVersion,
   pkgsAarch64,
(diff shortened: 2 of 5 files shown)


mikatammi:
https://github.com/mikatammi/nv-optee/commit/5251df6acc75495d6749fc9ce6cec0d39827b6f7

This could also be included as a patch to optee

mikatammi:
Okay, I got the TA loading working, I was able to load pcks11-ta from disk. The patches I added are here:

https://github.com/mikatammi/jetpack-nixos/tree/optee-load-path

mikatammi (approved):
Okay, this patch as is fine I think, but for building one's own TAs to be convenient, few additional changes are needed

mikatammi:
I created another PR #130 to add more patches to enable loading and building of TAs

PR #104, opened 2023-06-07
Add public key option for Optee OS runtime TA verification

jmbaur:
###### Description of changes

This allows for signature verification of runtime trusted applications that will override the default test key published in optee's source. Using this option requires that runtime TAs are signed using the corresponding private key. See
https://optee.readthedocs.io/en/latest/architecture/porting_guidelines.html#trusted-application-private-public-keypair for more information.

###### Testing

Tested on an orin agx devkit by:
1. loading a signed hello world TA and verifying it works
2. loading an unsigned version of the same TA and verifying it does not load into optee


modules/flash-script.nix
```diff
@@ -124,6 +124,17 @@ in
             type = types.listOf types.str;
             default = [];
           };
+
+          taPublicKey = mkOption {
```
danielfullmer, line 128: Could you rename this to `taPublicKeyFile` for consistency with the other options?

danielfullmer (changes requested):

danielfullmer:
TAs built in to the optee image do not need to be signed by this key, right?

modules/flash-script.nix
jmbaur: done!

jmbaur:
> TAs built in to the optee image do not need to be signed by this key, right?

Correct

danielfullmer (approved):

PR #115, opened 2023-07-28
Add patch for fixing missing EFI configuration tables on T194

jmbaur:
###### Description of changes

When all serial ports are disabled on a T194 device via a DTB overlay to the UEFI DTB, many EFI configuration tables end up missing (e.g. SMBIOS). This patch enables other configuration tables to still be setup properly in this scenario.

###### Testing

Tested on an xavier agx with all serial ports disabled in UEFI device tree.

### DIFF
--- a/uefi-firmware.nix
+++ b/uefi-firmware.nix
@@ -1,4 +1,4 @@
-{ lib, stdenv, buildPackages, fetchFromGitHub, runCommand, edk2, acpica-tools,
+{ lib, stdenv, buildPackages, fetchFromGitHub, fetchpatch, runCommand, edk2, acpica-tools,
   dtc, python3, bc, imagemagick, unixtools, applyPatches, nukeReferences,
   l4tVersion,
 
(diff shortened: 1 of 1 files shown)


uefi-firmware.nix
```diff
@@ -59,7 +59,13 @@ let
       rev = "2c81e0fc74f703012dd3b2f18da5be256e142fe3"; # Latest on r35.3.1-updates as of 2023-05-17
       sha256 = "sha256-Qh1g+8a7ZcFG4VmwH+xDix6dpZ881HaNRE/FJoaRljw=";
     };
-    patches = edk2NvidiaPatches ++ [ ./capsule-authentication.patch ];
+    patches = edk2NvidiaPatches ++ [
+      (fetchpatch {
+        url = "https://patch-diff.githubusercontent.com/raw/NVIDIA/edk2-nvidia/pull/68.patch";
```
danielfullmer, line 64: Recommend using a url like this one instead:
````
https://github.com/NVIDIA/edk2-nvidia/commit/9604259b0d11c049f6a3eb5365a3ae10cfb9e6d9.patch
```
The path-diff contents might change if the branch changes.

Also, inside the patch, would we not want to keep the DEBUG message?

danielfullmer (changes requested):

jmbaur:
I left out the debug message since the T234's behavior in this instance is to not print anything either.

uefi-firmware.nix
jmbaur: I left out the debug message since the T234's behavior in this instance is to not print anything either.

danielfullmer: Sounds good to me!

PR #117, opened 2023-07-30
Support for l4t Docker containers

### LINKED ISSUE #5: Enable support for nvidia-container
We should be able to run docker/podman and systemd containers (OCI) on jetson devices using nvidia's container runtime support.

j-baker:
Fixes #5. I'm very new to Nix and so would appreciate a code style review if that's possible. This is a little closer to 'prototype' than 'good to go' but I figured it'd be worth pushing up regardless as others will likely benefit!

###### Description of changes

After this MR, setting `virtualisation.docker.enableNvidia = true` will work on Jetsons. Unfortunately there are two other flags which need handling:

1. At time of writing the unified cgroup hierarchy is not supported by libnvidia-container, with the standard flags set. Therefore one must add `systemd.enableUnifiedCgroupHierarchy = false` or receive errors. I assume that with nixpkgs changes this will eventually fade.
2. One must also disable the assertion `assertion = cfg.enableNvidia -> config.hardware.opengl.driSupport32Bit or false` which with a big stick could be done by `assertions = lib.mkForce []`. I have a PR for this here: https://github.com/NixOS/nixpkgs/pull/246179 but obviously it will be quite some time before it shows up in stable channels.

###### Testing

Tested on orin-nano. Tested by starting l4t-base, apt installing cuda, compiling samples, and running them (with success results).

### DIFF
--- a/avoid-static-libtirpc-build.patch
+++ b/avoid-static-libtirpc-build.patch
@@ -1,0 +1,24 @@
+diff --git a/Makefile b/Makefile
+index 709102b..2c50cc6 100644
+--- a/Makefile
++++ b/Makefile
+@@ -201,7 +201,7 @@ $(LIB_STATIC_OBJ): $(LIB_OBJS)
+ 
+ all: CPPFLAGS += -DNDEBUG
+ all: STRIP  := @echo skipping: strip
+-all: shared static
++all: shared
+ 
+ # Run with ASAN_OPTIONS="protect_shadow_gap=0" to avoid CUDA OOM errors
+ debug: CFLAGS += -pedantic -fsanitize=undefined -fno-omit-frame-pointer -fno-common -fsanitize=address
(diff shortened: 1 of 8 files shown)


WatweA:
This PR merges master into your branch and has been tested on an Xavier NX - please take a look
https://github.com/j-baker/jetpack-nixos/pull/1

jmbaur:
@j-baker would you be able to rebase this PR? A few files in the project got restructured pretty recently.

jmbaur:
#172 is a successor to this PR. It uses NVIDIA's support for the CDI spec to declare what needs to be injected into the container. It also does not require disabling cgroups v2 on the host. It's been tested with a few different container images from nvcr.io/nvidia such as l4t-base and l4t-jetpack, but if there are any issues encountered, let us know!

PR #130, opened 2023-09-21
Don't add TEEC_TEST_LOAD_PATH into TA search paths

mikatammi:
###### Description of changes

This is to supplement PR #103 

 * By default, tee-supplicant prepends "/foo:/bar::/baz" into the TEEC_LOAD_PATH. Add a patch which removes that.
 * Export buildOpteeTaDevKit to make it easier to build TAs in other flakes

###### Testing

AGX devkit and NX devkit

### DIFF
--- a/0001-Don-t-prepend-foo-bar-baz-to-TEEC_LOAD_PATH.patch
+++ b/0001-Don-t-prepend-foo-bar-baz-to-TEEC_LOAD_PATH.patch
@@ -1,0 +1,36 @@
+From 0a29db5517155f9f6c0b756d61781512b1424a03 Mon Sep 17 00:00:00 2001
+From: Mika Tammi <mika.tammi@unikie.com>
+Date: Sat, 23 Sep 2023 00:37:42 +0300
+Subject: Don't prepend /foo:/bar::/baz to TEEC_LOAD_PATH
+
+Signed-off-by: Mika Tammi <mika.tammi@unikie.com>
+---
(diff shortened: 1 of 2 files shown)


jmbaur:
For that patch you have, wouldn't we just need to remove the hardcoded `#define TEEC_TEST_LOAD_PATH` that optee has and that should suffice? It would make the patch much easier to maintain going forward. It looks like CFG_TEST_TA_PATH already defaults to off.

jmbaur:
#132 exposes the built ta-dev-kit, let me know if this implementation would work for your use case.

mikatammi:
> For that patch you have, wouldn't we just need to remove the hardcoded `#define TEEC_TEST_LOAD_PATH` that optee has and that should suffice? It would make the patch much easier to maintain going forward. It looks like CFG_TEST_TA_PATH already defaults to off.

Actually this issue is present in the OP-TEE upstream (and not only NVIDIA's fork) and I opened an issue and PR there:
https://github.com/OP-TEE/optee_client/issues/357
https://github.com/OP-TEE/optee_client/pull/358

mikatammi:
But yes you are right, the proper fix is much more bigger

mikatammi:
Now I force-pushed a simplified version of this PR and the patch in it. Given that the #132 also gets merged the TA support is now convenient enough

jmbaur (approved):
This is great! I'm hoping to see your upstream patches to optee get merged as well, it would beneficial to have those backported against nvidia's port.

PR #132, opened 2023-09-21
Expose `taDevKit` in `config.system.build.jetsonDevicePkgs.taDevKit`

jmbaur:
###### Description of changes

    
The TA dev kit is specific per-board and it is necessary for this to be
exposed in order for users to build trusted applications that don't get
built into the OPTEE image.
    
This can be used when building a TA by using the make flag
"TA_DEV_KIT_DIR=${config.system.build.jetsonDevicePkgs.taDevKit}/export-ta_arm64".

###### Testing

Ensured `taDevKit` builds on a nixos system closure for an orin-agx-devkit


mikatammi:
This patch is better fix than my #130

Now one does not need to manually round-trip args back to the buildTaDevKit

mikatammi (approved):
Tested now. Works well 👍

PR #136, opened 2023-10-03
Build tee-supplicant with a customizable CFG_TEE_PLUGIN_LOAD_PATH

TanelDettenborn:
###### Description of changes
Adding a NixOS option for tee-supplicant to modify CFG_TEE_PLUGIN_LOAD_PATH by build time. As an example usage of setting CFG_TEE_PLUGIN_LOAD_PATH:

```
hardware.nvidia-jetpack.firmware.optee.suppPluginLoadPath = pkgs.linkFarm "optee-plugin-path" [
{
    name = "myTeeSupp.plugin";
    path = "myTeeSuppPluginDerivation";
}];
```

###### Testing
Tested on orin-nx-devkit

### DIFF
--- a/modules/flash-script.nix
+++ b/modules/flash-script.nix
@@ -158,6 +158,14 @@ in
             '';
           };
 
+          clientExtraMakeFlags = mkOption {
+            type = types.listOf types.str;
+            default = [];
+            description = lib.mdDoc ''
+              Extra make flags to pass OP-TEE's client build.
+           '';
+          };
+
           patches = mkOption {
             type = types.listOf types.path;
             default = [];
@@ -166,6 +174,9 @@ in
           extraMakeFlags = mkOption {
             type = types.listOf types.str;
             default = [];
+            description = lib.mdDoc ''
+              Extra make flags to pass OP-TEE's OS build.
+           '';
           };
 
           taPublicKeyFile = mkOption {
--- a/device-pkgs/default.nix
+++ b/device-pkgs/default.nix
@@ -36,7 +36,9 @@ let
 
   teeSupplicant = opteeClient.overrideAttrs (old: {
     pname = "tee-supplicant";
-    buildFlags = (old.buildFlags or []) ++ [ "CFG_TEE_CLIENT_LOAD_PATH=${cfg.firmware.optee.clientLoadPath}" ];
+    buildFlags = (old.buildFlags or [])
+      ++ [ "CFG_TEE_CLIENT_LOAD_PATH=${cfg.firmware.optee.clientLoadPath}" ]
+      ++ cfg.firmware.optee.clientExtraMakeFlags;
     # remove unneeded headers
     postInstall = ''
       rm -rf $out/include


modules/flash-script.nix
```diff
@@ -123,6 +123,15 @@ in
             '';
           };
 
+          suppPluginLoadPath = mkOption {
```
jmbaur, line 126:
```suggestion
          supplicantPluginLoadPath = mkOption {
```

jmbaur:
Is there a use case for dynamically changing plugins that exist under the tee supplicant plugin path? If not, I propose this option's type be `types.listOf types.package`. If empty, we can build tee supplicant with `CFG_TEE_SUPP_PLUGINS=n`, and if not empty, we can use `linkFarm` or something similar and build tee supplicant with `CFG_TEE_PLUGIN_LOAD_PATH=${accumulatedPlugins}` or whatever name we end up choosing. In that case, a name like `supplicantPlugins` probably works.

I recognize that the existing option `clientLoadPath` defaults to a path where it makes it easy for the user to dynamically load TAs at runtime, but I think that it should probably move to this style as well with just a list of derivations.

TanelDettenborn:
Maybe an explicit option for `CFG_TEE_SUPP_PLUGINS` would make more sense than deducting it from `CFG_TEE_PLUGIN_LOAD_PATH`? Maybe a potential user would like to keep plugin framework even if they would not create plugins during "initial deployment". Equally same would be if user would know that there are no use for plugins and it can be turned off.

Thinking further then adding an option for `CFG_TEE_PLUGIN_LOAD_PATH` might be excessive due same thing could be done with tee-supplicant command line parameters. Daemon could start with `plugin-path`-option.

TanelDettenborn:
As a rationale to this pull request then this is related to plugin framework and OPTEE xtests (with pkcs11 support).

I have a fork with xtests and pkcs11: https://github.com/TanelDettenborn/jetpack-nixos

TanelDettenborn:
@jmbaur I have rebased and updated pull request. Updated pull request addresses the same problem, but its approach is more common than a specific option. It introduces nixos option for providing extra make flags to OP-TEE client build. Even if the extra make flags makes clientLoadPath redundant then I have kept it for legacy reasons. 

For documentation purposes I have kept the original topic and body, but if preferred then I can update topic and body accordingly. Or if this feature is not desired then I can close this pull request.

jmbaur:
I might have been thinking about this incorrectly. `tee-supplicant` has a `--plugin-paths` cli argument that it will accept to override any default value specified at build time. We should probably make the default more nixos-friendly, but we do already have a nixos option for passing extra args to tee-supplicant.

TanelDettenborn:
Apologies for my vague expressions. It might have been clearer, if I would have edited pull request topic/body or closed and opened another pull request. 

This pull request introduces nixos-option for providing a list of extra make flags to OP-TEE's client build. It does not add a specific option for a specific flag, but a general option. Option type is a list of strings. It is a similar option as there is already for [tos-build](https://github.com/anduril/jetpack-nixos/blob/master/modules/flash-script.nix#L166). OP-TEE client can be configured with following flags: [config.mk](https://github.com/OP-TEE/optee_client/blob/master/config.mk). 

For example if there is a need for minimizing(/removing unused features) from OP-TEE client then currently it is not possible. For example GP socket support (CFG_GP_SOCKETS) can be only removed at the build time.

I do not have a strong opinions about user/nixos-friendliness, but kind of related then I have another pull request open ([pkcs11 and xtest](https://github.com/anduril/jetpack-nixos/pull/149)) in which `tee-supplicant` [CLI arguments are break into smaller parts](https://github.com/anduril/jetpack-nixos/pull/149/files#diff-826389d2e36b4094c8a2ab49d67b8d6ea42e488fba7b3c58c67ccd4a553011f8R131) for sake of code readability rather than user/nixos-friendliness.

jmbaur:
My opinion is that we should not continue adding NixOS module options for modifying how packages are built, and rather allow the user use `nixpkgs.overlays` to add their own build customizations. Many packages built by this repo are not currently in a condition to do this however (any package that ends up under `config.system.build.jetsonDevicePkgs`). In #156 I've placed `opteeClient` under the main package set so that it will end up under `pkgs.nvidia-jetpack.opteeClient`, so this should be extendable by just using regular overlays. I'm also working further to move more packages into a condition where they can be overridden more easily without adding further nixos module options.

So since #156 is not landed, I add my approval here, but I think what I've mentioned above is how we should move forward with similar changes.

jmbaur (approved):

TanelDettenborn:
Overlays would do it and using them might be a better approach for the project! I have skimmed through #156 and it is a solid work. As soon as OP-TEE's client patch is final and merged into OP-TEE main then this pull request will be outdated/useless. So rather than merging it into main, I would close this pull request to save from conflicts and keeping commit history clean.

jmbaur:
@danielfullmer I think this PR can be closed since it is resolved by #156

PR #137, opened 2023-10-03
Don't apply formatOptions unless systemd-makefs is not being used

### LINKED ISSUE #105: nixpkgs master no longer supports `fileSystems.<name>.formatOptions`
Attempting to use `jetpack-nixos` with nixpkgs master will fail this assertion:

https://github.com/NixOS/nixpkgs/blob/5d486408b510035c76b31d8d880c830d773b40a4/nixos/modules/tasks/filesystems.nix#L330-L332

It seems like it will be necessary to use some other mechanism to create the partition in question.

(Side note, the `/dev/disk/by-partlabel/esp` does not exist and is not mounted on my AGX anyhow. Not sure if this is just a step I missed when setting up `jetpack-nixos` or something else, though)

jmbaur:
###### Description of changes

With commit
https://github.com/nixos/nixpkgs/commit/5176a4f11356a338331d82e63563f8510b67317d, the option `fileSystems.<name>.options` will contain an "x-systemd.makefs". Since systemd-makefs does not take any parameters (but has sane defaults and actually uses fat32 for vfat fstype), the `formatOptions` option was made internal only and set to null. Let's set it null here unless systemd-makefs is not being used.

Fixes #105 

###### Testing

Tested that the assertion https://github.com/NixOS/nixpkgs/blob/28562b6c75906da822189e6f93bf45768dc04ee6/nixos/modules/tasks/filesystems.nix#L323 does not fail when evaluating a nixos config (using nixpkgs/nixos-unstable) with som xavier-agx and mountFirmwareEsp set to true. Also tested with nixpkgs/nixos-23.05 that formatOptions is a string with the value "-F 32 -n ESP".

### DIFF
--- a/modules/devices.nix
+++ b/modules/devices.nix
@@ -110,13 +110,19 @@ in lib.mkMerge [{
   # https://forums.developer.nvidia.com/t/setting-uefi-variables-using-the-defaultvariabledxe-only-works-if-esp-is-on-emmc-but-not-on-an-nvme-drive/250254
   # We don't mount this at /boot, because we still want to allow the user to have their ESP part on NVMe, or whatever else.
   fileSystems."/opt/nvidia/esp" = lib.mkDefault {
-    device = "/dev/disk/by-partlabel/esp";
-    fsType = "vfat";
-    options = [ "nofail" ];
-    # Since we have NO_ESP_IMG=1 while formatting, the script doesn't
-    # actually create an FS here, so we'll do it automatically
-    autoFormat = true;
-    formatOptions = "-F 32 -n ESP";
+    config = {
+      device = "/dev/disk/by-partlabel/esp";
+      fsType = "vfat";
+      options = [ "nofail" ];
+      # Since we have NO_ESP_IMG=1 while formatting, the script doesn't
+      # actually create an FS here, so we'll do it automatically
+      autoFormat = true;
+      formatOptions =
+        if (lib.versionAtLeast config.system.nixos.release "23.05") then
+          null
+        else
+          "-F 32 -n ESP";
+    };
   };
 })
 ]


modules/devices.nix
```diff
@@ -109,14 +109,17 @@ in lib.mkMerge [{
...
+  fileSystems."/opt/nvidia/esp" = lib.mkDefault ({ config, ... }: {
+    config = {
+      device = "/dev/disk/by-partlabel/esp";
+      fsType = "vfat";
+      options = [ "nofail" ];
+      # Since we have NO_ESP_IMG=1 while formatting, the script doesn't
+      # actually create an FS here, so we'll do it automatically
+      autoFormat = true;
+      # needed on nixpkgs < 23.11
+      formatOptions = if (lib.any (opt: opt == "x-systemd.makefs") config.options) then null else "-F 32 -n ESP";
```
danielfullmer, line 121: Is this just to condition on NixOS version? The better way would be to condition on `config.system.nixos.release`

danielfullmer (changes requested):

danielfullmer:
Does the UEFI capsule update test still pass on Xavier AGX?  The only thing we lose with this change is setting the filesystem label to "ESP", but I'm not sure whether the capsule update depends on that or not.

jmbaur:
I have not yet tested capsule updates with this change. From what I can see, systemd-makefs does not use a deterministic FS label.

jmbaur:
I verified that this change does not break capsule updates on Xavier devices. Since we reference the partition label and not the filesystem label, systemd-makefs would not influence any change in how we mount the partition.

danielfullmer:
Thanks!

PR #138, opened 2023-10-04
Wait for initrd to complete before initrd flash script exits

### LINKED ISSUE #84: Provide ability for initrd flash script to wait for completion
Currently it just outputs the following message on the host machine and exits:
```
Jetson device should now be flashing and will reboot when complete.
You may watch the progress of this on the device's serial port
```

jmbaur:
###### Description of changes

Use `expect` to watch the serial console of a device being flashed. This allows scripting of the initrd flash script based on exit code. If the flash fails, the script exits with code 1.

Other changes include fixing the debug shell.

Fixes #84 

###### Testing

Tested on an orin agx devkit


danielfullmer:
Negative on this implementation. I don't want to take over the serial console to wait for this, since other scripts may be watching serial console simultaneously.  We should instead be able to get this signal over the USB port somehow, since we know we have exclusive access to it.

PR #139, opened 2023-10-04
Add option for changing boot order

jmbaur:
###### Description of changes

This change adds a nixos option for changing the default boot order. The option defaults to using the same boot order from
https://github.com/NVIDIA/edk2-nvidia/blob/71fc2f6de48f3e9f01214b4e9464dd03620b876b/Silicon/NVIDIA/Tegra/DeviceTree/L4TConfiguration.dts, which gets applied in the flash script using `additionalDtbOverlays`.

###### Testing

Tested on an orin agx devkit

### DIFF
--- a/modules/flash-script.nix
+++ b/modules/flash-script.nix
@@ -24,6 +24,13 @@ in
       firmware = {
         autoUpdate = lib.mkEnableOption "automatic updates for Jetson firmware";
 
+        bootOrder = mkOption {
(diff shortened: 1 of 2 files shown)


danielfullmer:
I'd still like the ability to not set the default boot order via a device tree overlay and leave nvidia to do what it wants. How about changing the option to be nullOr (....), and if its null we don't apply any DTBO

jmbaur:
Changed to use `nullOr`!

PR #147, opened 2023-10-29
fix(initrdFlash): lib.getExe path to rcm-boot script (instead of dire…

johnrichardrinehart:
###### Description of changes

`writteShellApplication` places the executable in something like `$NIX_STORE_ROOT/<hash>-<name>-<version>/bin/<name>` (cf. https://github.com/NixOS/nixpkgs/blob/90e85bc7c1a6fc0760a94ace129d3a1c61c3d035/pkgs/build-support/trivial-builders/default.nix#L348) . The `initrdFlashScript` was apparently written/intended/designed to call the shell application defined by `mkRcmBootScript {... }` but it only pointed to `$NIX_STORE_ROOT/<hash>-<name>-<version>` (the `$out` folder instead of the executable).

We have used `lib.getExe` in this PR to resolve the path to the executable instead of to the folder.

###### Testing

We have used this patch to build and flashing script that points to an executable path which exists in our `nix` store :smile:  .

### DIFF
--- a/device-pkgs/default.nix
+++ b/device-pkgs/default.nix
@@ -183,11 +183,11 @@ let
   in writeShellApplication {
     name = "initrd-flash-${hostName}";
     text = ''
-      ${mkRcmBootScript {
+      ${lib.getExe (mkRcmBootScript {
         kernelPath = "${config.boot.kernelPackages.kernel}/Image";
         initrdPath = initrd;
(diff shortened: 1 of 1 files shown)


danielfullmer:
Thanks for looking into this. For consistency with the other `mk*` functions in that file, I'd rather we have that function just output text, and let the consumer of that function create the real shell script. Fixed in https://github.com/anduril/jetpack-nixos/pull/148

PR #165, opened 2023-12-01
uefi-firmware: fix edk2 build against NixOS 23.11

mikatammi:
###### Description of changes

uefi-firmware stopped building with NixOS 23.11, as util-linux (which includes libuuid) was removed from the depsBuildBuild of the edk2-package. Add libuuid to the depsBuildBuild, which then fixes the issue related to `-luuid`.

###### Testing

Tested by flashing on Orin AGX and Orin NX. Tested that the build works against NixOS 22.05, 23.05 and 23.11


therishidesai (approved):
Tested this as well and it works

jonringer (approved):
Did something similar in https://github.com/anduril/jetpack-nixos/pull/169

PR #173, opened 2023-12-15
treewide: format nix code and add CI to enforce formatting

jmbaur:
###### Description of changes

Open to discussions, but it would be nice to have some formatting applied :)

Also added is a `.git-blame-ignore-revs` file (also in nixpkgs) that can be used in a gitconfig to prevent `git blame` from showing specific commit revs. See https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt for more info.


jonringer (approved):
nixpkgs-fmt sounds good to me

danielfullmer:
Is my artisanally-crafted ad-hoc formatting not good enough for you, that you want to bring in a machine to do this? j/k
I'm in favor of something consistent--even if I don't always agree with nixpkgs-fmt.

jmbaur:
> Is my artisanally-crafted ad-hoc formatting not good enough for you, that you want to bring in a machine replace my job? j/k LGTM, I'm in favor of something consistent--even if I don't always agree with nixpkgs-fmt.

Hey I feel that! Hand-crafting some complex snippet of code just to have it all normalized by some silly formatting program feels weird! But yeah overall I agree with consistency.

PR #174, opened 2023-12-15
Allow for disabling tee-supplicant

jmbaur:
###### Description of changes

Add an `enable` option for tee-supplicant, but retain the current behavior of enabling it by default. This should remove some attack surface of devices that don't need tee-supplicant but have an optee instance running that is not configured to verify signatures of TAs at runtime. 

Also organize tee-supplicant related options into the `supplicant` attrset.

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -11,12 +11,12 @@ let
 
   teeApplications = pkgs.symlinkJoin {
     name = "tee-applications";
-    paths = cfg.firmware.optee.trustedApplications;
+    paths = cfg.firmware.optee.supplicant.trustedApplications;
   };
 
   supplicantPlugins = pkgs.symlinkJoin {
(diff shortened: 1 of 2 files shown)


modules/default.nix
```diff
@@ -204,15 +204,15 @@ in
       wantedBy = [ "multi-user.target" ];
     };
 
-    systemd.services.tee-supplicant =
+    systemd.services.tee-supplicant =
```
danielfullmer, line 207: Extra whitespace?

danielfullmer (changes requested):

danielfullmer:
Otherwise, PR looks OK to me.

modules/default.nix
jmbaur: removed!

Issue #171, opened 2023-12-12
Cannot run `nix build github:anduril/jetpack-nixos#iso_minimal`


tpwrules:
I get the following error on master (74880473fa4f9b69d97d81128de6437d6a068df2):
```
$ nix build github:anduril/jetpack-nixos#iso_minimal
error: cannot coerce null to a string

       at /nix/store/ikpqn4r16qds821k2fdxjp0mcdv7qj4d-source/lib/strings.nix:229:10:

          228|     # Input string
          229|     str: substring 0 (stringLength pref) str == pref;
             |          ^
          230|
(use '--show-trace' to show detailed location information)
```

Appears to be due to `hardware.jetpack-nixos.som` missing from the installer config. Not sure the best way to fix.

vilvo:
Depending on your need, quick workaround enables `iso_minimal` build:
```
+++ b/modules/default.nix
@@ -92,7 +92,8 @@ in

       # Needed on Orin at least, but upstream has it for both
       "nvidia.rm_firmware_active=all"
-    ] ++ lib.optional (lib.hasPrefix "xavier-" cfg.som) "video=efifb:off"; # Disable efifb driver, which crashes Xavier AGX/NX
+      "video=efifb:off"
+    ]; #++ lib.optional (lib.hasPrefix "xavier-" cfg.som) "video=efifb:off"; # Disable efifb driver, which crashes Xavier AGX/NX
```
I only need this to install NixOS on AGX and I want that crashing framebuffer out.

PR #175, opened 2023-12-15
Fix eval: Rework some option types to not be nullOr

danielfullmer:
###### Description of changes

We were implicitly using the "null" value on the som and carrierBoard options as a way to specify that the values were unset and that jetpack-nixos should try to maximize compatibility acroos all types (with caveats that this may not always include all features). Now, we just have a separate "generic" value that we can use to represent this. It's easier to work with since we don't need to have "!= null" checks everywhere.

###### Testing

`nix flake check` for eval also ran internal graphics tests successfully .

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -51,18 +51,23 @@ in
       };
 
       som = mkOption {
-        default = null;
         # You can add your own som or carrierBoard by merging the enum type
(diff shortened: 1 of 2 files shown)


jmbaur:
I think there's another cfg.som null check in `device-pkgs/default.nix`

jmbaur:
Closes #171

PR #183, opened 2024-01-24
[kernel] Fix memory leak in kernel tegra serial driver

kevinchristensen1:
###### Description of changes
Fixes a memory leak in kernel tegra serial driver.
This ensures the memory allocated in function `tegra_dma_prep_slave_sg`
is cleaned up properly. Manifested as slab unreclaimable growing unbounded
via calls to kmalloc-256

Tested on Xavier AGX custom carrier board

Ref: https://forums.developer.nvidia.com/t/serial-tegra-dma-driver-bug/258168/5

### DIFF
--- a/kernel/default.nix
+++ b/kernel/default.nix
@@ -1,6 +1,7 @@
 { pkgs
 , lib
 , fetchFromGitHub
+, fetchpatch
 , l4t-xusb-firmware
 , realtime ? false
 , kernelPatches ? [ ]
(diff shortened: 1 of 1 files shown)


kernel/default.nix
```diff
@@ -91,6 +91,11 @@ pkgsAarch64.buildLinux (args // {
 
     # Fix gcc13 compilation failure
     { patch = ./0009-bonding-gcc13-synchronize-bond_-a-t-lb_xmit-types.patch; }
+    
+    # Fixes a memory leak by kernel tegra serial driver
+    # This manifested via slab unreclaimable growing unbounded via repeated kmalloc-256 calls
+    # Patch is applied on tegra-gpc-dma.c
+    { patch = ./0010-tegra-gpc-dma-mem-leak.patch; }
```
jonringer, line 98: if we know the patch version which contains the upstream fixes, we can conditionally add this. Otherwise someone will need to edit this later.

kevinchristensen1: I believe we only want to apply this for `35.3.X` since @danielfullmer mentioned `35.4.1` has this patch already 

Any suggestions for how to do the conditional apply?

employee-64c7dcd530593118dcccc3fb: If there's a way to detect that we're past `d5b90d6b9365250adb73b2fe5b52a5228df3b1d9` in https://github.com/OE4T/linux-tegra-5.10/commits/l4t/l4t-r35.4.ga-5.10/ then we can do it. Looks like it's at [7191dccf8670635906182cc2da862d9c0fdcb93a](https://github.com/anduril/jetpack-nixos/blob/23645d7e360d6158a91f514a63deea362493005d/kernel/default.nix#L28), now.

Although, I think it's okay to add it unconditionally and leave a comment that it can be removed for any version past `d5b90d6b9365250adb73b2fe5b52a5228df3b1d9`. There are a number of patches, here, and any of them might also need conditional application. Easiest just to document when they can be popped off, even if it would break if a user overrides the `rev` they want to use.

danielfullmer: No need to apply conditionally. I'll just remove the patch when I do the upgrade to 35.4.1 (which should hopefully happen soon)

kevinchristensen1: Awesome sounds good guys, thanks!

danielfullmer (approved):

PR #187, opened 2024-02-07
Stable link to OE4T/meta-tegra instead of master

IlyaNiklyaev:
Recent update in repo https://github.com/OE4T/meta-tegra (this one OE4T/meta-tegra@b86c2ae) in fact breaks `jetpack-nixos` build:
```
hash mismatch in fixed-output derivation '/nix/store/n59rw7dri9dacvrav379m3i621qql6nf-0003-Update-conversion-defaults-to-match-NVIDIA-sources.patch.drv':
      >          specified: sha256-vGilgHWinrKjX+ikHo0J20PL713+w+lv46dBgfdvsZM=
      >             got:    sha256-6xCEjf432FLksCCk8+EpAbQU3PeGICOQ+6Q2OusUNa4=
```

I think we should refer all external repos by stable links to the certain commits instead of branches.


danielfullmer (approved):

danielfullmer:
Thanks!  You're right, all references to patches on github should refer explicitly to a commit and not to a branch.

PR #190, opened 2024-03-11
Add `coreutils` to device package scripts

watwea-heaviside:
###### Description of changes

Add `coreutils` to device scripts (i.e. flash script, initrd flash, etc.). Because the tools in this package were missing, a packaged version of the script (built for 3rd party use outside of a `nix`-enabled machine) using `nix bundle .#SCRIPT_NAME` resulted in errors like: 
```
...initrd-flash-orin-nx-devkit: line 9: mktemp: No such file or directory
```

###### Testing

rebuilding with the `coreutils` package on an x86_64-linux system results in successfully running the script packaged with `nix bundle --out-link ~/script .#SCRIPT_NAME; ~/script` without the missing file error above

### DIFF
--- a/device-pkgs/default.nix
+++ b/device-pkgs/default.nix
@@ -16,6 +16,7 @@
 , dtc
 , l4tVersion
 , pkgsAarch64
+, coreutils
 ,
 }:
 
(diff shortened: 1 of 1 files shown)


danielfullmer:
So, coreutils is already present in flashDeps, do regular flash scripts work with `nix bundle` and is it just an issue with initrdFlashScript?

PR #198, opened 2024-03-28
Update to Jetpack 5.1.3 / L4T 35.5.0

Princemachiavelli:
###### Description of changes
Jetpack release notes: https://docs.nvidia.com/jetson/archives/jetpack-archived/jetpack-513/release-notes/index.html
Major release notes are "Double-Bit ECC Error Detection for the Jetson AGX Orin Industrial" and "Fixes for known security vulnerabilities.".

However, a major undocumented change is edk2-nvidia has upgraded from OpenSSL 1.1.1 to using 3.0. This PR removes `vendoredOpenSSL` and instead follows upstream nixpkg's approach of using `buildPackages.openssl.src`.

#### Notes
* Upstream edk2-nvidia NVIDIA/edk2-nvidia#76 to fix the Eqos TX clock name appears to be unresolved; vendoring the patch here until fixed in upstream.
* Upstream linux-tegra broke DMI_SYSFS by disabling DMI itself (OE4T/linux-tegra-5.10@bc94634fcddd594735aa9c5ca5f68b4df1cb5f8b) for some unknown bug; this PR reverts this change to re-enable DMI and DMI_SYSFS.
* Upstream started using `chipsku` in addition to `boardsku` to differentiate Orin NX and Orin Nano so this needs to be tracked under firmware variants.
 
 
### Updating
- [x] Update `l4tVersion`, `jetpackVersion`, and `cudaVersion` in default.nix
- [x] Update branch/revision/sha256s in:
    - [x] `default.nix`
    - [x] `kernel/default.nix`
    - [x] `uefi-firmware.nix`
    - [x] Grep for "sha256 = " and "hash = ", see if there is anything else not covered
- [x] Update gitrepos.json using sourceinfo/gitrepos-update.py  and result/source_sync.sh from bspSrc.
- [x] Update the kernel version in `kernel/default.nix` if it chaged.
- [x] Grep for the previous version strings e.g. "35.5.0"
- [x] Compare files from `unpackedDebs` before and after
- [x] Grep for NvOsLibraryLoad in libraries from debs to see if any new packages not already handled in l4t use the function
- [x] Ensure the soc variants in `modules/flash-script.nix` match those in `jetson_board_spec.cfg` from BSP
- [x] Ensure logic in `ota-utils/ota_helpers.func` matches `nvidia-l4t-init/opt/nvidia/nv-l4t-bootloader-config.sh`
- [x] Run `nix build .#genL4tJson` and copy output to `pkgs/containers/l4t.json`

### Testing
- [x] Run `nix flake check`
- [x] Build installer ISO
- [x] Flash all variants
- [x] Boot all variants
- [x] Test UEFI Capsule update (35.3.1 -> 35.5.0)
- [x] Run our (Anduril's) internal automated device tests


modules/flash-script.nix
```diff
@@ -198,9 +198,12 @@ in
         eksFile = mkOption {
           type = types.nullOr types.path;
           default = null;
+          description = "
```
danielfullmer, line 201: This should be in a different PR.
Also, it would need to be triple quotes, instead of double quotes.

pkgs/samples/default.nix
```diff
@@ -229,6 +229,7 @@ let
 
     unpackCmd = "dpkg -x $src source";
     sourceRoot = "source/usr/src/jetson_multimedia_api";
+    separateDebugInfo = true;
```
danielfullmer, line 232: This isn't a valid option for upstream nixpkgs.

pkgs/uefi-firmware/default.nix
```diff
@@ -73,28 +73,21 @@ let
...
-        url = "https://github.com/NVIDIA/edk2-nvidia/commit/9604259b0d11c049f6a3eb5365a3ae10cfb9e6d9.patch";
-        hash = "sha256-v/WEwcSNjBXeN0eXVzzl31dn6mq78wIm0u5lW1jGcdE=";
-      })
       # Fix Eqos driver to use correct TX clock name
       # PR: https://github.com/NVIDIA/edk2-nvidia/pull/76
-      (fetchpatch {
-        url = "https://github.com/NVIDIA/edk2-nvidia/commit/26f50dc3f0f041d20352d1656851c77f43c7238e.patch";
-        hash = "sha256-cc+eGLFHZ6JQQix1VWe/UOkGunAzPb8jM9SXa9ScIn8=";
-      })
+      # Can't find commit from Nvidia that actually fixes PR#76
```
danielfullmer, line 82: Want to remove this new comment? They still haven't fixed it upstream. If you need to vendor the patch like you've done, that's fine, but could you also include the commit message?

danielfullmer (changes requested):
Thanks for your work on this so far!

pkgs/uefi-firmware/default.nix
Princemachiavelli: Comment removed and commit messaged added to patch. Thanks.

pkgs/samples/default.nix
jmbaur: I think it is now: https://github.com/nixos/nixpkgs/blob/9306e3cc87e0bcf686ededc6060add6d3245cca2/doc/stdenv/stdenv.chapter.md?plain=1#L1088

Princemachiavelli:
Should be ready for review again now.
Besides resolving all the issues identified, there are a few changes.

1. Added `chip_sku` to `hardware.nvidia-jetpack.firmware.variants`, this functions just like `boardsku` when flashing except now the flash script needs `chip_sku` to be externally set in order to correctly differentiate between Orin NX and Orin Nano.
2. nvfancontrol: noticed this service would sometimes start too early and fail so enable `Restart=on-failure`.
3. optee-gen-ekb: Since regenerating the EKB is necessary for fused boards, I've provided the tool as a standalone package.
4. optee-ftpm-manufacturer: Not strictly necessary for this PR but needed for using the new fTPM feature.

danielfullmer:
Could you fix the formatting CI failures and then I'd be ready to merge. Thanks!

Princemachiavelli:
@danielfullmer This is ready for review again. It passes our internal device tests. The CI failure appears to be a GitHub timeout issue because it succeeds locally for me and `nixpkgs-fmt --check` also passes.

danielfullmer:
To be superseded by https://github.com/anduril/jetpack-nixos/pull/254.  Thanks for your work on this, as it was very helpful to refer to during the update process.

PR #201, opened 2024-04-01
flashDeps: use buildPackages to fix cross-builds

Princemachiavelli:
###### Issue

Cross builds of nixosConfig containing an UEFI capsule update are currently broken due to how `flashDeps` are defined and used.

```
nix-repl> :b (nixosConfigurations.installer_minimal_cross.extendModules {
    hardware.nvidia-jetpack.enable = true; 
    modules = [{
      hardware.nvidia-jetpack.firmware.autoUpdate = true;
      hardware.nvidia-jetpack.som = "orin-nx";
     }];
}).config.system.build.toplevel
```

Results in exec format error due to wrong package set of `flashDeps` being used.
```
/build/.attr-0l2nkwhif96f51f4amnlf414lhl4rv9vh8iffyp431v6s28gsr90: line 24: /nix/store/g6s3cb7y0mmfa5i1s7xr5j9ma58rcwwn-coreutils-aarch64-unknown-linux-gnu-9.3/bin/cp: cannot execute binary file: Exec format error
/build/.attr-0l2nkwhif96f51f4amnlf414lhl4rv9vh8iffyp431v6s28gsr90: line 6: /nix/store/g6s3cb7y0mmfa5i1s7xr5j9ma58rcwwn-coreutils-aarch64-unknown-linux-gnu-9.3/bin/rm: cannot execute binary file: Exec format error
``` 

###### Changes
Use `buildPackages` for `flashDeps` so `flash-script.nix` uses the correct package set. As far as I can tell, there's no way to change which package set callPackage uses for passthru so directly using `buildPackages` seems like the best option.

The use of `pkgs.buildPackages.callPackage` is required for `patchShebangs`  to set the correct interpreter in `flash.sh`. (Using `callPackage` for some reason just resulted in the shebang being left unmodified). There might be some other cause for this issue.

###### Testing
TODO

### DIFF
--- a/modules/devices.nix
+++ b/modules/devices.nix
@@ -35,9 +35,11 @@ let
     (lib.concatLines (map
       # all files in linux-firmware are read-only
       (firmwarePath: ''
-        install -Dm0444 \
-          --target-directory=$(dirname $out/lib/firmware/${firmwarePath}) \
-          $(realpath ${pkgs.linux-firmware}/lib/firmware/${firmwarePath})
+        if [ -f ${pkgs.linux-firmware}/lib/firmware/${firmwarePath} ]; then
+          install -Dm0444 \
+            --target-directory=$(dirname $out/lib/firmware/${firmwarePath}) \
+            $(realpath ${pkgs.linux-firmware}/lib/firmware/${firmwarePath})
+        fi
       ''
       )
       paths)));
--- a/default.nix
+++ b/default.nix
@@ -76,7 +76,7 @@ let
     inherit bspSrc gitRepos l4tVersion;
   }) buildTOS buildOpteeTaDevKit opteeClient;
 
-  flash-tools = callPackage ./pkgs/flash-tools {
+  flash-tools = pkgs.buildPackages.callPackage ./pkgs/flash-tools {
     inherit bspSrc l4tVersion;
   };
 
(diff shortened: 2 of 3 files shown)


modules/devices.nix
```diff
@@ -35,9 +35,11 @@ let
     (lib.concatLines (map
       # all files in linux-firmware are read-only
       (firmwarePath: ''
-        install -Dm0444 \
-          --target-directory=$(dirname $out/lib/firmware/${firmwarePath}) \
-          $(realpath ${pkgs.linux-firmware}/lib/firmware/${firmwarePath})
+        if [ -f ${pkgs.linux-firmware}/lib/firmware/${firmwarePath} ]; then
```
jmbaur, line 38: Some of the files in linux-firmware are symlinks, I don't think `-f`still works in that case. Can we also pull this change into a separate PR?

jmbaur:
> The use of pkgs.buildPackages.callPackage is required for patchShebangs

Where does this `patchShebangs` happen? I can't find it

modules/devices.nix
Princemachiavelli: Yea, the change is just so I can test this PR. I intend to remove it.

Princemachiavelli:
> > The use of pkgs.buildPackages.callPackage is required for patchShebangs
> 
> Where does this `patchShebangs` happen? I can't find it

I think it's added to the fixup in `pkgs/build-support/setup-hooks/patch-shebangs.sh`? I still don't understand why it's behavior would differ between `pkgs.callPackage` and `pkgs.buildPackages` unless the way it checks for build time commands fails if the architecture is different.

PR #204, opened 2024-04-16
fix libv4l patch sha

rprince-anduril:
###### Description of changes
I guess github changed their patch format? This was failing with a FOD hash mismatch on nixbuild testing where we didn't have a cached version.

###### Testing
Tested in anduril-nixpkgs


danielfullmer:
Thanks, superseded by https://github.com/anduril/jetpack-nixos/pull/205 which uses fetchurl instead and fixes another mistake fetching patches

PR #207, opened 2024-04-20
Don't require using systemd-boot to get automated firmware updates

jmbaur:
###### Description of changes

Using systemd-boot is not a requirement for applying capsule updates. We can make the firmware update a systemd service, which gives us the behavior of running on every `switch-to-configuration` when the capsule update file changes.

###### Testing

Tested that the capsule update was successfully applied on an Orin AGX Devkit and a fused Xavier AGX Devkit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ Weston and sway have been tested working on Orin devices, but do not work on Xav
 ### Updating firmware from device
 Recent versions of Jetpack (>=5.1) support updating the device firmware from the device using the UEFI Capsule update mechanism.
 This can be done as a more convenient alternative to physically attaching to the device and re-running the flash script.
-These updates can be performed automatically after a `nixos-rebuild boot` if the `hardware.nvidia-jetpack.bootloader.autoUpdate` setting is set to true and systemd-boot is used.
+These updates can be performed automatically after a `nixos-rebuild switch` if the `hardware.nvidia-jetpack.bootloader.autoUpdate` setting is set to true.
 Otherwise, the instructions to apply the update manually are below.
 
 To determine if the currently running firmware matches the software, run, `ota-check-firmware`:
--- a/modules/flash-script.nix
+++ b/modules/flash-script.nix
@@ -1,4 +1,4 @@
-{ config, pkgs, lib, ... }:
+{ config, pkgs, lib, utils, ... }:
 
 # Convenience package that allows you to set options for the flash script using the NixOS module system.
 # You could do the overrides yourself if you'd prefer.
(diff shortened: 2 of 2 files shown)


danielfullmer:
My concern with this is that the current behavior feels more correct (except for the fact that it's tied to systemd-boot). Currently, if you do a `switch-to-configuration test`, it does not try to apply a UEFI update, but if you do a `switch-to-configuration boot` or `switch-to-configuration switch` then it will apply the update.  There's similar behavior in for systemd-boot updates in NixOS, where the systemd-boot update would only happen if you did a `boot` or `switch`.  Unfortunately, I don't see another option that is not systemd specific to additionally do this update at the same time.

jmbaur:
I've modified the change to not apply the capsule update if `nixos-rebuild test` is used. Surprisingly, there is little logic in `switch-to-configuration` regarding when `test` is the subcommand, only that the bootloader install script is not ran.

danielfullmer:
I think that should be sufficient. Does this PR pass our uefi update test?

jmbaur:
> I think that should be sufficient. Does this PR pass our uefi update test?

Yes they do! With and without auto update enabled

danielfullmer:
@jmbaur  I believe this PR caused a regression in update behavior. Say we are upgrading from version X to version Y:

Previously, running `nixos-rebuild boot` against a NixOS system containing newer firmware Y (and with `autoUpdate` enabled), would drop the capsule update in the right folder on the ESP so that the firmware would pick it up and apply it during the next reboot reboot (before loading the new NixOS system).  That would ensure that after a reboot, both the firmware and the NixOS system would be on the new version Y.

Now, because the update logic has been moved entirely into the `firmware-update` service, when running `nixos-rebuild boot`, nothing would get dropped into ESP, and during the first reboot, no capsule update would be applied.  So after the first reboot it'd be running firmware version X with NixOS version Y.  Then, after booting up the `firmware-update` service finally runs, so on the next reboot it'll come up with both firmware version and NixOS version Y.

I'm not confident that NVIDIA provides any guarantees for anything working if firmware and system software are mismatched.  It hasn't been a problem for 35.3.1 -> 35.4.1 or 35.4.1 -> 35.6.0, but I don't want to depend on that going forward.

README.md
```diff
@@ -129,7 +129,7 @@ Weston and sway have been tested working on Orin devices, but do not work on Xav
 ### Updating firmware from device
 Recent versions of Jetpack (>=5.1) support updating the device firmware from the device using the UEFI Capsule update mechanism.
 This can be done as a more convenient alternative to physically attaching to the device and re-running the flash script.
-These updates can be performed automatically after a `nixos-rebuild boot` if the `hardware.nvidia-jetpack.bootloader.autoUpdate` setting is set to true and systemd-boot is used.
+These updates can be performed automatically after a `nixos-rebuild switch` if the `hardware.nvidia-jetpack.bootloader.autoUpdate` setting is set to true.
```
danielfullmer, line 132: I should have notice this earlier.  We don't want to run `nixos-rebuild switch` when upgrading major versions of Jetpack/L4T, since then the new version of the system would be running a different kernel version, etc.

PR #209, opened 2024-04-26
feat: add support for flashing emulation configs on AGX devkits

therishidesai:
###### Description of changes

[The AGX supports Emulating weaker Jetson SoMs on the devkit.](https://developer.ridgerun.com/wiki/index.php/NVIDIA_Jetson_Orin/Flashing_commands_for_emulation#Flashing_the_board). This PR add's additional AGX devkit options that can be flashed to enable this emulation. 

###### Testing

Tested this on an Orin AGX devkit.

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -80,10 +80,7 @@ in
       };
 
       carrierBoard = mkOption {
-        type = types.enum [
-          "generic"
-          "devkit"
-        ];
+        type = types.enum [ "generic" "devkit" "devkit-as-nx-8gb" "devkit-as-nx-16gb" "devkit-as-nano-8gb" "devkit-as-nano-4gb" ];
         default = "generic";
         description = ''
           Jetson carrier board to target. Can be set to "generic" to target a generic jetson carrier board, but some things may not work.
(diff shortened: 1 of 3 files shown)


therishidesai:
@jmbaur @danielfullmer can we get some reviews on this? Would like to have these features in upstream instead of a separate fork

modules/default.nix
```diff
@@ -80,10 +80,7 @@ in
       };
 
       carrierBoard = mkOption {
-        type = types.enum [
-          "generic"
-          "devkit"
-        ];
+        type = types.enum [ "generic" "devkit" "devkit-as-nx-8gb" "devkit-as-nx-16gb" "devkit-as-nano-8gb" "devkit-as-nano-4gb" ];
```
elliotberman, line 83: It's not really proper to make this a characteristic of the carrierBoard, since it seems to be more a property of the SOM.

We've currently have support for "super" mode for the nx/nano. What if we deprecated the super option and introduce new option: `somPower` (exact name up to you) which is an enum that supports "default", "super", "nx-8gb", "nx-16gb", etc.

elliotberman (changes requested):
Hi @therishidesai , had discussed your PR recently and providing some late feedback, if you're still interested.

PR #212, opened 2024-05-02
Package l4t-gstreamer

matshch:
###### Description of changes
Added nvidia-l4t-gstreamer package, which provides several GStreamer plugins.

###### Testing
Ran GStreamer pipeline with several NVidia plugins on AGX Orin:
```bash
gst-launch-1.0 videotestsrc ! video/x-raw,width=176,height=144 ! nvvidconv ! nvv4l2h264enc ! h264parse ! qtmux ! filesink location=a.mp4
```

### DIFF
--- a/pkgs/l4t/default.nix
+++ b/pkgs/l4t/default.nix
@@ -217,6 +217,11 @@ let
     '';
   };
 
+  l4t-gstreamer = buildFromDeb {
+    name = "nvidia-l4t-gstreamer";
+    buildInputs = [ l4t-camera l4t-cuda l4t-multimedia wayland ];
+  };
+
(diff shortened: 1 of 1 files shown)


danielfullmer:
Confirmed that the gstreamer plugins appear to be working.  In the future, we can see if its possible to build these from source using the repositories [here](https://github.com/anduril/jetpack-nixos/blob/2790dd8153bc1bda30258adae3b3353a31d9c28c/sourceinfo/r35.4.1-gitrepos.json#L254-L397) . Additionally, we could add some gstreamer tests to the multimedia test.  Merging as-is, since this does seem to work.

PR #223, opened 2024-06-25
Deprecate nixos options related to packaging

jmbaur:
###### Description of changes

We have a few nixos options that are mirrors of arguments to `stdenv.mkDerivation` (e.g. `patches`, `postPatch`). Let's encourage users to use the overlay system in order customize packaging.


danielfullmer:
Instead of simply removing these options, I think we should throw a warning when they are used, but remove them entirely at some later date.  Optimally, we would also have documented an example of a way to do what one of these options did, but using an overlay.

jmbaur:
> Instead of simply removing these options, I think we should throw a warning when they are used, but remove them entirely at some later date. Optimally, we would also have documented an example of a way to do what one of these options did, but using an overlay.

Added warnings here to warn of future deprecation: https://github.com/anduril/jetpack-nixos/pull/223/files#diff-826389d2e36b4094c8a2ab49d67b8d6ea42e488fba7b3c58c67ccd4a553011f8R357

PR #228, opened 2024-07-11
l4t-multimedia: Fix a data race in shutting down dequeue thread

hacker1024:
###### Description of changes

There's a bug in the multimedia API sample classes that leads to a segmentation fault in a few cases when using a non-Debian glibc.

The PR adds [OE4T's patch](https://github.com/OE4T/meta-tegra/blob/master/recipes-multimedia/argus/tegra-mmapi-samples/0004-samples-classes-fix-a-data-race-in-shutting-down-deq.patch) to fix the issue.


danielfullmer:
Thanks!

PR #233, opened 2024-08-02
Add option for postFlashCommands

jmbaur:
###### Description of changes

The argument to `flash-script.nix` already existed, this just adds an option that allows users to hook into this.

###### Testing

Tested building the flash script with a non-empty value set in the new option and ensured the value was present in the final script.


jmbaur:
I actually didn't end up needing this option. Probably best not to add it unless a user needs it, as it becomes hard to remove options once they exist :)

PR #244, opened 2024-09-23
Add bc to nv_fuse_read.sh path

jmbaur:
###### Description of changes

The nv_fuse_read.sh shell script (in `l4t-tools`) expects `bc` to be in PATH.

###### Testing

Tested on a fused xavier-agx and an unfused orin-agx.


danielfullmer:
Looks pretty unobjectionable.

PR #245, opened 2024-09-24
Move more packages into the overlay

jmbaur:
###### Description of changes

This change adds more packages from `device-pkgs` into the overlay that contributes to the system's package-set. The packages moved over in this change are notably ones that have the `requiredSystemFeatures` attribute. This means that we could eventually remove the `requiredSystemFeatures` option and let users change the derivation how they need in their own overlay.

###### Testing

Tested flash, initrd flash, UEFI capsule update, and basic boot test on an orin-agx-devkit and a fused xavier-agx-devkit

### DIFF
--- a/device-pkgs/default.nix
+++ b/device-pkgs/default.nix
@@ -8,12 +8,7 @@
 { lib
 , dtc
 , gcc
-, makeInitrd
 , nvidia-jetpack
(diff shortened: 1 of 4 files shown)


danielfullmer:
:crossed_fingers:

PR #250, opened 2024-09-30
Add console.enable option to modules/default.nix

brian-jordan:
###### Description of changes
Added console.enable option to allow for disabling the default console boot kernelParams

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -125,7 +125,8 @@ in
       };
 
       console.enable = mkOption {
-        default = true;
+        default = config.console.enable;
+        defaultText = "config.console.enable";
         type = types.bool;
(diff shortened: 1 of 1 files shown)


modules/default.nix
```diff
@@ -123,6 +123,12 @@ in
           binaries needed for flashing/fusing Jetson SOMs.
         '';
       };
+
+      console.enable = mkOption {
+        default = true;
```
jmbaur, line 128:
```suggestion
        default = config.console.enable;
        defaultText = "config.console.enable";
```
 There exists an option upstream that might be nice to mirror that value here.

jmbaur (changes requested):

PR #251, opened 2024-10-16
nvidia-ctk: 1.15.0 -> 1.16.3

jmbaur:
###### Description of changes

This update also drops the patches needed to build nvidia-ctk with go < v1.20, as this version is no longer in any supported nixos release.

###### Testing

OCI test in `pkgs/tests/default.nix` runs successfully on an xavier-agx and orin-agx (both devkit).


danielfullmer:
You love to see it

PR #254, opened 2024-11-18
Update to Jetpack 5.1.4 / L4T 35.6.0

danielfullmer:
###### Description of changes

Jetpack 5.1.4 release notes: https://docs.nvidia.com/jetson/archives/jetpack-archived/jetpack-514/release-notes/index.html
L4T 35.6.0 release notes: https://docs.nvidia.com/jetson/archives/r35.6.0/ReleaseNotes/Jetson_Linux_Release_Notes_r35.6.0.pdf

###### Testing

- [x] Run nix flake check
- [x] Build installer ISO
- [x] Flash all variants
- [x] Boot all variants
- [x] Test UEFI Capsule update (35.3.1 -> 35.5.0)
- [x] Run our (Anduril's) internal automated device tests


pkgs/uefi-firmware/edk2-uefi-dtb.patch
```diff
@@ -1,15 +1,14 @@
...
-                   NULL,
-                   &gEfiEventReadyToBootGuid,
-                   &ReadyToBootEvent
+@@ -561,8 +561,7 @@ InstallFdt (
+   }
+ 
+   PlatformType = TegraGetPlatform ();
+-  if ((PlatformType != TEGRA_PLATFORM_SILICON) ||
+-      (PlatformResourceInfo->BootType == TegrablBootRcm))
++  if (TRUE) // Always use DTB from UEFI
```
hacker1024, line 11: Why was this change made? Does this prevent the [`hardware.deviceTree`](https://search.nixos.org/options?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=hardware.deviceTree) options from working? We use it to merge the kernel-supplied device tree with camera overlays.

danielfullmer: The original patch no longer applied since upstream refactored this area of the codebase.  Unlike upstream's edk2 builds, we don't use default to using the `kernel-dtb` partitions to load/store the device tree for the Linux kernel, but instead always re-use the device tree included with the uefi firmware.  The change I made here shouldn't change behavior from how it worked previously.  `hardware.deviceTree` options should still work.

danielfullmer:
Finally merging this update.

Note that the UEFI variable store will be reinitialized after a capsule update to JP5.1.4 due to the new variable authentication feature.  A funny (and just sad) consequence of this is that Secure Boot is disabled if it was enabled imperatively through the UEFI menu, instead of declaratively autoenrolling keys through the device tree.  See: https://forums.developer.nvidia.com/t/uefi-secure-boot-is-canceled-when-capsule-update-from-r35-4-1-to-r35-6-0/317302

Additionally, this reinitialization also appears to be the reason `nvbootctrl dump-slots-info` does not report a successful update, as mentioned [here](https://github.com/anduril/jetpack-nixos/pull/198#issuecomment-2146229768).  The ESRT variable LastAttemptVersion does not persist across the reboot, so `nvbootctrl` doesn't think an update actually happened.  CC @Princemachiavelli

PR #263, opened 2025-01-06
Fix source URLs(?)

elikoga:
###### Description of changes

Looks like nvidia changed their source URLs by adding a `/r/` path segment. I can build `.#iso_minimal` using the current change.


danielfullmer:
I ran `git clone` against a few of the original URLs without the `/r/` and they were able to check out successfully for me.  Do you have a specific example of one that fails?

elikoga:
I cannot reproduce the failure that prompted me to make this change currently. I suspect this may be user error or an intermittent reconfiguration on nvidias side.
Thank you for investigating this though

PR #264, opened 2025-01-06
Allow `nixos-rebuild boot` to work with automated firmware updates

jmbaur:
###### Description of changes

In 9cea389171151c28dab47dfc1b08401e87aabfa0, we broke the ability for using `nixos-rebuild boot` to apply capsule updates. This change adds that ability back. The systemd service is kept as we still want capsule updates to be applied in cases where a different bootloader is being used (or perhaps some other niche setup).

###### Testing

Manually tested on an xavier-agx-devkit:
- `nixos-rebuild boot` and `nixos-rebuild switch` installs the capsule update if no prior one exists
- `nixos-rebuild test` does nothing

### DIFF
--- a/modules/flash-script.nix
+++ b/modules/flash-script.nix
@@ -9,6 +9,49 @@ let
     types;
 
   cfg = config.hardware.nvidia-jetpack;
+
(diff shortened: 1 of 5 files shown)


danielfullmer:
Thanks for this change!

PR #265, opened 2025-01-10
Fix platform firmware and UEFI capsule builds when cross compiling

jmbaur:
###### Description of changes

We must ensure we are using `flash-tools` for the build platform when we are using it during a build (i.e. not on a machine performing the flashing at "runtime").

This change fixes the immediate issue of some of the `packages` flake outputs not building (since we cross compile them), however there is some good future work that can be done to re-organize these derivations where we can just `callPackage` them and automatically get the right `flash-tools`.

###### Testing

Tested against internal test suite as well as tested the fixed flake outputs for uefi capsule updating and initrd flashing for an orin-nx devkit.

### DIFF
--- a/overlay-with-config.nix
+++ b/overlay-with-config.nix
@@ -144,7 +144,7 @@ final: prev: (
         }
         (''
           ${cfg.firmware.uefi.capsuleAuthentication.preSignCommands final.buildPackages}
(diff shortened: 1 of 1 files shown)


danielfullmer:
Makes sense. Thanks for the fix!

PR #266, opened 2025-01-13
nv-display-driver: GCC14 fix implicit function

brianmcgillion:
With the latest changes to gcc, introduced in gcc14 there conftest is failing. GCC14 is now part of the nixos-unstable so this is a preventative change and does not introduce a regression on gcc < 14.

###### Description of changes

do not enforce the strict checking of implicit-function-declaration

###### Testing

built against nixos-unstable


danielfullmer:
Thanks!

PR #267, opened 2025-01-15
uefi-firmware: Fix GCC 14 compile issue

TanelDettenborn:
UEFI-firmware compile fails with "missing binary operator before token "("", because "__has_builtin"-operator is not defined within gcc 14. Patching edk2-sources with existing patch. Patch is fetched from NVIDIA EDK2 repo.

###### Description of changes
Fetched patch checks if  "__has_builtin"-operator has been defined and then uses it. Further information:
*  GNU CPP [documentation](https://gcc.gnu.org/onlinedocs/gcc-14.2.0/cpp/_005f_005fhas_005fbuiltin.html) about "__has_builtin"-operator.
* tianocore/edk2 [pull request](https://github.com/tianocore/edk2/pull/5781)

###### Testing
Compiled with gcc 14 and  gcc 13 and both compiles.

### DIFF
--- a/pkgs/uefi-firmware/default.nix
+++ b/pkgs/uefi-firmware/default.nix
@@ -46,12 +46,22 @@ let
   ###
 
   # See: https://github.com/NVIDIA/edk2-edkrepo-manifest/blob/main/edk2-nvidia/Jetson/NVIDIAJetsonManifest.xml
-  edk2-src = fetchFromGitHub {
-    owner = "NVIDIA";
-    repo = "edk2";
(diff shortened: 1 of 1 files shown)


danielfullmer:
Looks reasonable. Could you correct the typo "uefi-firmaware" to "uefi-firmware" in the commit title and message?

TanelDettenborn:
Typo fixed in the commit title and messate.

danielfullmer:
Thanks!

PR #269, opened 2025-01-20
unused module errors

brianmcgillion:
###### Description of changes

https://github.com/NixOS/nixpkgs/pull/366004 introduced a change that
warns and errors out when modules are defined and not acutally set for a
particular target.

This patch rectifies that by going back the the previous method of
ignoring those errors and not breaking the build. Commentary is left
inline to explain the situation and possible future solutions.

tested against orin agx target for ghaf.

there are similar solutions being deployed in nixpkgs to work around
this e.g.

https://github.com/NixOS/nixpkgs/pull/375165

### DIFF
--- a/kernel/default.nix
+++ b/kernel/default.nix
@@ -16,6 +16,19 @@ buildLinux (args // {
 
   defconfig = "tegra_defconfig";
 
+  # https://github.com/NixOS/nixpkgs/pull/366004
+  # introduced a breaking change that if a module is declared but it is not being used it will fail
+  # if you try to suppress each of he errors e.g.
+  # REISERFS_FS_SECURITY = lib.mkForce unset; within structuredExtraConfig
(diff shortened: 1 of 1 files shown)


kernel/default.nix
```diff
@@ -99,9 +112,14 @@ buildLinux (args // {
     # Override the default CMA_SIZE_MBYTES=32M setting in common-config.nix with the default from tegra_defconfig
     # Otherwise, nvidia's driver craps out
     CMA_SIZE_MBYTES = lib.mkForce (freeform "64");
+    MODVERSIONS = no;
```
hacker1024, line 115: What's this for?

brianmcgillion: it was a left over while trying to track down the mismatch of module versions when building against mainline.

danielfullmer:
LGTM, thanks for the contribution!

PR #271, opened 2025-02-05
Use kernelPackages scope for nvidia-display-driver callPackage

hacker1024:
###### Description of changes

The `nvidia-display-manager` is not actually `callPackage`d from the `kernelPackages` scope, meaning that changes to packages within the scope (like the kernel itself) are not used in the `nvidia-display-manager` derivation.

In particular, this means that when using options like [`boot.kernelPatches`](https://search.nixos.org/options?channel=unstable&show=boot.kernelPatches), `nvidia-display-manager` erroneously still references the stock kernel, leading to two kernels being built.

This MR fixes this.

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -99,14 +99,14 @@ in
     tests = prev.callPackages ./pkgs/tests { inherit l4tVersion; };
 
     kernelPackagesOverlay = final: prev: {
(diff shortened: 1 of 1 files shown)


danielfullmer (approved):

danielfullmer:
Good catch! Thank you!

PR #277, opened 2025-02-14
Skip firmware updates if script is run on a virtualised system

### LINKED ISSUE #276: Bootloader-style automatic firmware updates are incompatible with make-disk-image.nix
Nixpkgs's [make-disk-image.nix](https://github.com/NixOS/nixpkgs/blob/master/nixos/lib/make-disk-image.nix) function, used for generating NixOS disk images, attempts to [install the bootloader while running in a VM](https://github.com/NixOS/nixpkgs/blob/6a00fac9cbd2be254b9396eee031b8620e6c21b3/nixos/lib/make-disk-image.nix#L726).

When `hardware.nvidia-jetpack.firmware.autoUpdate` is enabled, it adds the [firmware update script](https://github.com/anduril/jetpack-nixos/blob/5db499eebff8e15c4edc26ea14078423fc7a4803/modules/flash-script.nix#L15-L53) to the bootloader installation command list. This isn't suitable for a VM, though, for the following reasons:

- `/sys/devices/virtual/dmi/id/bios_version` doesn't exist, causing the update script (and transitively, the bootloader installation and disk image build) to `exit 1`
  - If the QEMU/UEFI setup in the VM changes in the future, it may start existing too, but its value will not be useful
- NVRAM variables to configure the capsule update don't actually get set on the Jetson

To resolve these issues, there are at least three different solutions:

1. `exit 0` instead of `exit 1` if `/sys/devices/virtual/dmi/id/bios_version` doesn't exist during bootloader installation. I don't think a failed firmware update attempt should necessarily indicate that the bootloader didn't get installed.
2. Work out somehow if a VM is in use, and don't attempt to update at all.
3. Work out somehow if a VM is in use, and update unconditionally. This will work as long as [`touchEFIVars`](https://github.com/NixOS/nixpkgs/blob/6a00fac9cbd2be254b9396eee031b8620e6c21b3/nixos/lib/make-disk-image.nix#L131) is used to set the NVRAM on the Jetson.

jmbaur:
###### Description of changes

We should not attempt to perform a firmware update if the system is being virtualised (e.g. in make-disk-image.nix).

Fixes #276

###### Testing

Tested to work with the following expression:

```nix
let
  jetpack-nixos = builtins.getFlake "github:jmbaur/jetpack-nixos/firmware-update-detect-virt";
  nixpkgs = builtins.getFlake "github:nixos/nixpkgs/nixos-24.11";
in
(nixpkgs.lib.nixosSystem {
  modules = [
    (
      {
        config,
        lib,
        pkgs,
        ...
      }:
      {
        imports = [ jetpack-nixos.nixosModules.default ];

        fileSystems."/".device = "/dev/foo";

        boot.loader.systemd-boot.enable = true;

        hardware.nvidia-jetpack = {
          enable = true;
          som = "orin-agx";
          carrierBoard = "devkit";
        };

        system.build.diskImage = import "${pkgs.path}/nixos/lib/make-disk-image.nix" {
          inherit config lib pkgs;
          label = "nixos";
          format = "qcow2";
          partitionTableType = "efi";
          touchEFIVars = true;
          installBootLoader = true;
        };
      }
    )
  ];
}).config.system.build.diskImage
```

Also tested that capsule updates on real hardware still work.

### DIFF
--- a/modules/flash-script.nix
+++ b/modules/flash-script.nix
@@ -14,8 +14,16 @@ let
 
   updateFirmware = pkgs.writeShellApplication {
     name = "update-jetson-firmware";
-    runtimeInputs = with pkgs; [ coreutils nvidia-jetpack.otaUtils ];
+    runtimeInputs = [ pkgs.coreutils config.systemd.package pkgs.nvidia-jetpack.otaUtils ];
     text = ''
+      # If this script is not run on real hardware, don't attempt to perform an
+      # update. This script could potentially run in a few places, for example
+      # in <nixpkgs/nixos/lib/make-disk-image.nix>.
+      if systemd-detect-virt --quiet; then
+        echo "virtualisation detected, skipping jetson firmware update"
+        exit 0
+      fi
+
       # This directory is populated by ota-apply-capsule-update, don't run if
       # we already have a capsule update present on the ESP. We check the exact
       # path that we populate because it is possible for multiple capsule


modules/flash-script.nix
```diff
@@ -14,8 +14,16 @@ let
 
   updateFirmware = pkgs.writeShellApplication {
     name = "update-jetson-firmware";
-    runtimeInputs = [ pkgs.nvidia-jetpack.otaUtils ];
+    runtimeInputs = [ config.systemd.package pkgs.nvidia-jetpack.otaUtils ];
     text = ''
+      # If this script is not ran on real hardware, don't attempt to perform an
+      # update. This script could potentially run in a few places, for example
+      # in <nixpkgs/nixos/lib/make-disk-image.nix>.
+      if [[ $(systemd-detect-virt) != "none" ]]; then
```
hacker1024, line 22: > `systemd-detect-virt` exits with a return value of 0 (success) if a virtualization technology is detected, and non-zero (error) otherwise.

https://www.freedesktop.org/software/systemd/man/latest/systemd-detect-virt.html#Description

```suggestion
      if systemd-detect-virt --quiet; then
```

modules/flash-script.nix
```diff
@@ -14,8 +14,16 @@ let
 
   updateFirmware = pkgs.writeShellApplication {
     name = "update-jetson-firmware";
-    runtimeInputs = [ pkgs.nvidia-jetpack.otaUtils ];
+    runtimeInputs = [ config.systemd.package pkgs.nvidia-jetpack.otaUtils ];
     text = ''
+      # If this script is not ran on real hardware, don't attempt to perform an
```
danielfullmer, line 19: Grammar: s/ran/run/
Also in commit title.

danielfullmer (changes requested):
Have you tested the regular UEFI capsule update workflow with this change?

PR #278, opened 2025-02-19
Orin AGX 64gb fab from "000" to "300"

TanelDettenborn:
###### Description of changes
Jetson 35.6 BSP "jetson_board_spec.cfg"-file uses/defines Orin AGX 64gb "300" as a fab value. 

###### Testing
Evaluates and compiles.


danielfullmer:
Thanks for the catch. It looks like this changed upstream between Jetpack 5.1.1 and Jetpack 5.1.2 in the `jetson_board_spec.cfg` file.  Since the TegraCompatSpec variable doesn't include the FAB for Orin AGX (see `ota_helpers.func`), and there doesn't appear to be any flash script behavior conditional on FAB for Orin AGX, I think this shouldn't have any functional difference. Still good to merge to keep in sync with Jetpack.

PR #284, opened 2025-03-27
cudaPackages: align with upstream

ConnorBaker:
#### Organization

This PR also moves to a pattern of scope creation and population through `packagesFromDirectoryRecursive`, relying on the contents of the filesystem rather than monolithic files with many package expressions inline. This is done for `cudaPackages` and `samples`.

#### Modules

The modules provided enable CUDA support and specify CUDA capabilities for the Xavier and Orin through `nixpkgs.config`.

#### Flake

Make `legacyPackages` use Nixpkgs configured with CUDA support for Xavier and Orin, as well as overlays which make the CUDA package set constructed within this repo the default.

Additionally, image construction is switched to `nixpkgs.lib.nixosSystem` from `pkgs.nixos` so the Nixpkgs instance used by the image is built from the config specified by the updated modules. (With `pkgs.nixos`, `nixpkgs.config` isn't used.)

#### cudaPackages

This PR refactors `cudaPackages` to use a redist-builder-like approach (similar to upstream), though the packages are still constructed from debian installers. The goal is to align with upstream's CUDA package sets, since it reduces the amount of breakage users of these packages face (missing outputs, setup hooks, etc.).

The benefits of this change (and alignment more generally with upstream) include:

- splayed outputs for reduced closure size
- the introduction of `flags` for parity with upstream's CUDA package sets
- the introduction of setup hooks from upstream and https://github.com/ConnorBaker/cuda-packages/tree/3fdad92cf320274cf24c87a30e3498a5d9fad18e which enable CMake to find and use CUDA dependencies

### Testing

- [x] nix flake check
- [x] build iso_minimal on both architectures
- [x] build our internal systems based on this
- [x] build and run our internal tests based on this


pkgs/cuda-packages/setupCudaHook/setup-cuda-hook.sh
```diff
@@ -167,8 +167,8 @@ disallowNVCCHostCompilerLinkLeakage() {
     while IFS= read -r -d ':' rpath; do
       for forbiddenRPATH in "${cudaForbiddenRPATHs[@]}"; do
         if [[ $rpath == "$forbiddenRPATH"* ]]; then
-          echo "ERROR: forbidden path $forbiddenRPATH exists in RPATH of $libpath"
-          return 1
+          nixErrorLog "ERROR: forbidden path $forbiddenRPATH exists in RPATH of $libpath"
+          # return 1 # Don't error because we don't have a mechanism to fix this at the moment.
```
danielfullmer, line 171: Add a clear note here that this is a differences from what's in upstream.

ConnorBaker:
For things which were vendored, add the commit in the file before the block which was copied.

When there are differences from upstream, make sure to call them out.

Split the samples commit into two -- one to move to a scope, and then one with fixes.

danielfullmer (approved):
Conditionally approved, assuming you document what has been tested.  We should, at least, test:
- [ ] nix flake check
- [ ]  build `iso_minimal` on both architectures
- [ ] build our internal systems based on this
- [ ] build and run our internal tests based on this

pkgs/cuda-packages/setupCudaHook/setup-cuda-hook.sh
ConnorBaker: Fixed in latest (the scanning for forbidden RPATH entries was removed, but differences are noted at the top of the file).

overlay.nix
```diff
@@ -76,29 +92,105 @@ in
...
+    python-jetson = final.python3.pkgs.callPackage ./pkgs/python-jetson { };
 
-    tegra-eeprom-tool = prev.callPackage ./pkgs/tegra-eeprom-tool { };
-    tegra-eeprom-tool-static = prev.pkgsStatic.callPackage ./pkgs/tegra-eeprom-tool { };
+    tegra-eeprom-tool = final.callPackage ./pkgs/tegra-eeprom-tool { };
+    tegra-eeprom-tool-static = final.pkgsStatic.callPackage ./pkgs/tegra-eeprom-tool { };
 
-    cudaPackages = prev.callPackages ./pkgs/cuda-packages {
-      inherit (self) debs cudaVersion
+    cudaPackages = makeScope self.newScope (finalCudaPackages: {
```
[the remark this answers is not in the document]
ConnorBaker: Moved `cudaConfig` to top-level of `pkgs/cuda-packages` to reduce inline size.

overlay.nix
```diff
@@ -76,29 +92,105 @@ in
...
+      inherit (self)
+        cudaVersion
+        debs
         l4t-3d-core
```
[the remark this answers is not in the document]
ConnorBaker: Removed it -- if it's needed, they can still be retrieved from the enclosing scope or `l4t`.

flake.nix
```diff
@@ -92,7 +92,27 @@
...
+            self.overlays.default
+            (final: prev: {
+              # NOTE: samples (and other packages) may pull in dependencies which depend on CUDA (either directly or
+              # transitively) -- this is problematic for us, because the default CUDA package set is not the one we
+              # construct.
+              # To avoid mixed package sets, we make our CUDA package set the default.
+              cudaPackages = final.nvidia-jetpack.cudaPackages;
+              # Reset OpenCV's override on cudaPackages present in 24.11.
+              # https://github.com/NixOS/nixpkgs/blob/7ffe0edc685f14b8c635e3d6591b0bbb97365e6c/pkgs/top-level/all-packages.nix#L10540-L10541
+              opencv4 = prev.opencv4.override { inherit (final) cudaPackages; };
```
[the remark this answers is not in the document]
ConnorBaker: Fixed in latest.

README.md
```diff
@@ -1,21 +1,21 @@
...
- * Platform firmware flashing scripts
- * A 5.10 Linux kernel from NVIDIA, which includes some open-source drivers like nvgpu
- * An [EDK2-based UEFI firmware](https://github.com/NVIDIA/edk2-nvidia)
- * ARM Trusted Firmware / OP-TEE
- * Additional packages for:
-   - GPU computing: CUDA, CuDNN, TensorRT
-   - Multimedia: hardware accelerated encoding/decoding with V4L2 and gstreamer plugins
-   - Graphics: Wayland, GBM, EGL, Vulkan
-   - Power/fan control: nvpmodel, nvfancontrol
+This repository packages components from NVIDIA's [ Jetpack SDK ] (https://developer.nvidia.com/embedded/jetpack) for use with NixOS, including:
```
danielfullmer, line 3: Can you remove these formatting changes?

README.md
```diff
@@ -199,7 +199,54 @@ CDI specification, so Podman is recommended for running containers on Jetson
 devices. Docker is set to get experimental CDI support in their version 25
 release.
 
+## Configuring CUDA for Nixpkgs
+
```
danielfullmer, line 203: Can you explain what the default behavior of using the NixOS module is, and mention the new configureCuda option up front?

If the specifics of the cudaSupport / cudaCapabilities is documented upstream already, you can just have a link to that documentation.

danielfullmer (changes requested):

README.md
ConnorBaker: Fixed in latest

README.md
ConnorBaker: Added a note about the default behavior.

There’s also no docs on this upstream, so I guess this is the first place these guidelines have been documented.

ConnorBaker:
- [x] nix flake check
- [x] build iso_minimal on both architectures
- [x] build our internal systems based on this
- [ ] build and run our internal tests based on this

danielfullmer (approved):

PR #289, opened 2025-05-05
Refactoring for Jetpack 6

elliotberman:
###### Description of changes

This series prepares the project for supporting jetpack 6. We'll simultaneously support jetpack 5 and jetpack 6. The UEFI firmware and kernel need unique derivations with minimal re-use between them, so move them to folders (r35 and tegra_5_10, respectively).

###### Testing

- [x] `nix flake check`


ConnorBaker:
@elliotberman Since `overlay.nix` is now a function which returns an overlay, I'd recommend renaming it to something like `mk-overlay.nix` or `generic-overlay.nix`, and creating a `jetpack5-overlay.nix` and `jetpack6-overlay.nix` which just contain something like (for JetPack 5):

```nix
import ./mk-overlay.nix {
  # Below 3 are from release notes
  jetpackVersion = "5.1.5";
  l4tVersion = "35.6.1";
  cudaMajorMinorPatchVersion = "11.4.298";

  bspHash = "sha256-nqKEd3R7MJXuec3Q4odDJ9SNTUD1FyluWg/SeeptbUE=";
}
```

This pattern of generic builder and minimal wrapper has worked well for CUDA stuff -- the files which provide a specific version (like the proposed `jetpack5-overlay.nix`) are the only place where that version information exists (a single source of truth).

In `mk-overlay.nix`, I'd want to see those first three versions included at the top-level of the `nvidia-jetpack` scope. (As a nit, I strongly prefer that version strings include, as part of the name, the components the version provides. For example, `jetpackVersion` would become `jetpackMajorMinorPatchVersion` and so on -- I've burned myself by using `versionOlder` and `versionAtLeast` with too few components!)

Other than that, looks good so far!

elliotberman:
> @elliotberman is there a public place where the refactoring is getting planned out? Would love to help out with the effort to support Jetpack6 in NixOS.

There are no public plans as such. I welcome any comments in this PR!

ConnorBaker:
Given the number of places you’re using `lib.versions`, it might make sense to add top level attributes for the versions you care about with different numbers of components (e.g., major, major and minor) and pass them in as callPackage-provided arguments rather than manually recreate them everywhere.

PR #290, opened 2025-05-08
Fix initrd flash script with Jetpack 5.1.5

danielfullmer:
###### Description of changes

Jetpack 5.1.5 changed how RCM (recovery mode) booting happened, which broke Xavier AGX and any fused/securebooting device.   This PR fixes those two issues.

###### Testing

Tested initrdFlashScript on Xavier AGX and on a fused Orin Nano.

### DIFF
--- a/modules/flash-script.nix
+++ b/modules/flash-script.nix
@@ -142,6 +142,25 @@ in
                 The path to the UEFI DB Signature List (ESL).
               '';
             };
+
+            signer.cert = mkOption {
+              type = lib.types.path;
+              description = ''
+                x509 certificate used for signing.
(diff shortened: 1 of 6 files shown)


modules/flash-script.nix
```diff
@@ -142,6 +142,25 @@ in
...
+              '';
+            };
+            signer.key = mkOption {
+              type = lib.types.path;
+              description = ''
+                RSA Private key in PEM format used for signing.
+                This is currently only used for RCM payload verification used in initrd flash script.
+                Besides this specific option, UEFI secure boot behaves as it normally would for any other NixOS system.
+                Do _not_ include a signing key in the Nix store. How to do this
+                is left as an exercise to the reader.
```
elliotberman, line 160: Could we do a build-time (run-time) test whether the key is in the nix store and bail or complain about it?

PR #294, opened 2025-05-28
Fix broken symlinks

elliotberman:
###### Description of changes

We had a few broken symlinks in cuda packages. The symlinks should only be created if the destination exists.

### DIFF
--- a/pkgs/cuda-packages/deb-builder/fixup.nix
+++ b/pkgs/cuda-packages/deb-builder/fixup.nix
@@ -176,8 +176,10 @@ in
         echo "removing existing symlink $PWD/lib64"
         rm "$PWD/lib64"
       fi
-      echo "symlinking $PWD/lib64 -> $PWD/lib"
-      ln -rs "$PWD/lib" "$PWD/lib64"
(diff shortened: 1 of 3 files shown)


pkgs/cuda-packages/deb-builder/fixup.nix
```diff
@@ -176,8 +176,10 @@ in
         echo "removing existing symlink $PWD/lib64"
         rm "$PWD/lib64"
       fi
-      echo "symlinking $PWD/lib64 -> $PWD/lib"
-      ln -rs "$PWD/lib" "$PWD/lib64"
+      if [ -n "$(find "$PWD/lib" -not \( -path "$PWD/lib/stubs" -prune \) -name \*.so)" ] ; then
```
ConnorBaker, line 179:
```suggestion
      if [[ -n "$(find "$PWD/lib" -not \( -path "$PWD/lib/stubs" -prune \) -name \*.so)" ]]; then
```

ConnorBaker (approved):
Elliot and I both feel leaving the TODO is appropriate -- it's unclear whether this branch will be taken in the future, but across many JetPack releases it hasn't been the case.

PR #298, opened 2025-06-18
optee: uefi-firmware: Compile/use standalonemm_optee from our edk2repo

elliotberman:
NVIDIA's recommended fix for "an assertion issue in UEFI during boot" requires recompiling StandaloneMMOptee. Add an additional derivation to build the StandaloneMmOptee platform and package it into OP-TEE.

Link: https://forums.developer.nvidia.com/t/assertion-issue-in-uefi-during-boot/315628
Fixes: a751cf1e23cd ("uefi-firmware: re-add patches fixing assertion errors")


pkgs/uefi-firmware/default.nix
```diff
@@ -159,11 +159,12 @@ let
...
   buildTarget = if debugMode then "DEBUG" else "RELEASE";
 
-  jetson-edk2-uefi =
+  mkJetsonUefi =
+    platform:
     # TODO: edk2.mkDerivation doesn't have a way to override the edk version used!
     # Make it not via passthru ?
     stdenv.mkDerivation (finalAttrs: {
-      pname = "jetson-edk2-uefi";
+      pname = "jetson-edk2-uefi-${platform}";
```
danielfullmer: How about
`pname = "${platform}-edk2-uefi"` instead?

overlay.nix
```diff
@@ -82,7 +82,7 @@ in
       # Nvidia's recommended toolchain is gcc9:
       # https://nv-tegra.nvidia.com/r/gitweb?p=tegra/optee-src/nv-optee.git;a=blob;f=optee/atf_and_optee_README.txt;h=591edda3d4ec96997e054ebd21fc8326983d3464;hb=5ac2ab218ba9116f1df4a0bb5092b1f6d810e8f7#l33
       stdenv = final.gcc9Stdenv;
-      inherit (self) bspSrc gitRepos l4tVersion;
+      inherit (self) bspSrc gitRepos l4tVersion uefi-firmware;
```
danielfullmer, line 85: `bspSrc` can be removed now?

pkgs/uefi-firmware/default.nix
danielfullmer: Also, is `platform` the right name for it? Not `builtTarget` or something?  We'll want to match whatever edk2/nvidia calls it--if it has a name for it.

pkgs/uefi-firmware/default.nix
```diff
@@ -252,6 +256,10 @@ let
       ${jetson-edk2-uefi}/AARCH64/L4TLauncher.efi \
       $out/L4TLauncher.efi
 
+    python3 ${edk2-nvidia}/Silicon/NVIDIA/Tools/FormatUefiBinary.py \
+      ${jetson-edk-uefi-stmm-optee}/FV/UEFI_MM.Fv \
+      $out/standalonemm_optee.bin
+
```
danielfullmer, line 262: Do we need to nuke-refs for this target like we do for uefi_jetson?

pkgs/uefi-firmware/default.nix
elliotberman: > Also, is platform the right name for it? Not builtTarget or something? We'll want to match whatever edk2/nvidia calls it--if it has a name for it.

I pondered this for a moment, too. We're referencing platform description files, so it felt okay to name it `platform`.

https://tianocore-docs.github.io/edk2-DscSpecification/release-1.28/#edk-ii-platform-description-dsc-file-specification

overlay.nix
elliotberman: Still referenced here:

https://github.com/anduril/jetpack-nixos/blob/d8280d7d1b000780546d813281d6a1ee97e952e3/pkgs/optee/default.nix#L247

pkgs/uefi-firmware/default.nix
elliotberman: Done.

pkgs/uefi-firmware/default.nix
danielfullmer: Got it, thanks!

PR #299, opened 2025-06-19
uefi-firmware: update version to include unique hash

### LINKED ISSUE #275: Automatic firmware updates do not get applied if the version does not change
The automatic firmware update script only runs [if the new firmware version is different to the existing one](https://github.com/anduril/jetpack-nixos/blob/d8d927bc052678c0ae0054bb73fb01a954091732/modules/flash-script.nix#L39).

Firmware can be changed without the version changing, though (for example, through the [patch options](https://github.com/anduril/jetpack-nixos/blob/5db499eebff8e15c4edc26ea14078423fc7a4803/modules/flash-script.nix#L101-L115)).

It'd be great if there was a way to handle that situation appropriately too. Maybe the firmware derivation or output path hash can be added to the version string?

danielfullmer:
###### Description of changes

Currently, the version field we bake into the UEFI firmware is just the L4T version.  A simple version field does not any of the additional customizations (such as patches) we or the end-user may do on top of it. This makes it difficult to determine if the current firmware on a device includes the customizations expected.

Additionally, since the firmware autoupgrade feature keys off of the version number, a user has to know to manually apply a capsule update if they changed something that affects the platform firmware, which is not always obvious.

This adds a hash to the version field, like: `35.6.1-06cb4270ebfe`. Now, firmware should automatically be upgraded if there is a change that can affect platform firmware. This gets us a level of "congruence" similar to NixOS itself.  The downside is that _any_ change which affects the derivation hashes included in the firmware build will cause an upgrade, such as minor bumps to nixpkgs which cause mass-rebuilds. These upgrades can take approximately 4 minutes during reboot.  Overall, I believe the tradeoff is worth it.

Fixes #275

Co-authored-by: Elliot Berman <eberman@anduril.com>

###### Testing

Manually tested on an Orin NX devkit that the hash is included in the firmware, and that `ota-check-firmware` shows the current and expected firmware version match.
Also tested flashing, booting, and capsule updating across all 4 variants.

### DIFF
--- a/overlay-with-config.nix
+++ b/overlay-with-config.nix
@@ -44,6 +44,22 @@ final: prev: (
         errorLevelInfo = cfg.firmware.uefi.errorLevelInfo;
         edk2NvidiaPatches = cfg.firmware.uefi.edk2NvidiaPatches;
         edk2UefiPatches = cfg.firmware.uefi.edk2UefiPatches;
+
+        # A hash of something that represents everything that goes into the
+        # platform firmware so that we can include it in the firmware version.
+        # BUP should include everything relevant. However, bup also includes a
+        # reference to uefi-firmware which would cause infinite recursion while
+        # calculating the derivation hashes, so we need to "quotient out" the
+        # uefi-firmware.  We do a fancy override scope here to make a version
+        # of the bup that is otherwise identical, but does not depend on
+        # uefi-firmware. The level of magic here can be frightening.
+        uniqueHash =
+          let
+            cursedBup = (finalJetpack.overrideScope (a: b: {
+              uefi-firmware = null;
+            })).bup;
+          in
+          builtins.hashString "sha256" "${cursedBup}";
       } // lib.optionalAttrs cfg.firmware.uefi.capsuleAuthentication.enable {
         inherit (cfg.firmware.uefi.capsuleAuthentication) trustedPublicCertPemFile;
       });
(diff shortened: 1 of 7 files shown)


overlay-with-config.nix
```diff
@@ -44,6 +44,20 @@ final: prev: (
...
+        # reference to uefi-firmware which would cause infinite recursion while
+        # calculating the derivation hashes, so we need to "quotient out" the
+        # uefi-firmware.  We do a fancy override scope here to make a version
+        # of the bup that is otherwise identical, but does not depend on
+        # uefi-firmware. The level of magic here can be frightening.
+        uniqueHash = let
+          cursedBup = (finalJetpack.overrideScope (a: b: {
+            uefi-firmware = null;
+          })).bup;
+        in builtins.hashString "sha256" (builtins.unsafeDiscardStringContext cursedBup.drvPath);
```
elliotberman: Since we set `uefi-firmware = null;` above, I wonder whether we need to use unsafeDiscardStringContext here or in the other spot we use it? I don't think there'd be a circular dependency.

danielfullmer: Yeah we can remove unsafeDiscardStringContext entirely I believe, since hashString wouldn't retain a reference to any string context of the stuff it hashed.

elliotberman: `cursedCode--;` :tada:

PR #301, opened 2025-06-20
Ignore existing capsule file

jmbaur:
###### Description of changes

We don't actually care if an existing capsule file is present on the ESP when we are applying the capsule built with a NixOS configuration. We actually _do_ want to always apply the capsule tied to the current NixOS configuration, since it may be needed for booting that configuration's kernel or other components (e.g. a NixOS config has a Jetpack 6 kernel and a capsule update to Jetpack 6 firmware, however some other Jetpack 5.x capsule file is already present).

###### Testing

TODO

### DIFF
--- a/modules/flash-script.nix
+++ b/modules/flash-script.nix
@@ -24,16 +24,6 @@ let
         exit 0
       fi
 
-      # This directory is populated by ota-apply-capsule-update, don't run if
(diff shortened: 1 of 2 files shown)


danielfullmer (approved):
Conditionally approved, assuming your testing works fine.

PR #306, opened 2025-06-25
Switch to upstream nvidia-container-runtime

elliotberman:
###### Description of changes

The upstream nvidia-container-runtime now supports the CSV file format that Jetpack uses and we can switch to use that.

###### Testing

I found this in my `reflog`, need to make sure it works.


elliotberman:
Upstream doesn't actually (yet) support what's needed for Jetson devices. Holding on this for now.

pperanich:
Is this simply waiting for the backport to be merged into the release branch, or was the fix within not sufficient to enable support on Jetson devices?

PR #309, opened 2025-06-27
nvidia-container-toolkit: Migrate hardware.nvidia-jetpack.container-toolkit.enable

elliotberman:
###### Description of changes

virtualisation.{docker,podman}.enableNvidia has been deprecated by upstream nixpkgs (nixos/nixpkgs#284507). The upstream nvidia-container-toolkit is also not yet suitable for us (nixos/nixpkgs#344729). Create our own config to gate the generation of the CDI configuration.

###### Testing

- [x] Run `oci-test`
- [x] Play with enabling `virtualisation.docker.enableNvidia`, `hardware.nvidia-jetpack.container-toolkit.enable`, `hardware.nvidia-container-toolkit.enable` to observe the warnings/assertions


danielfullmer (approved):

modules/default.nix
```diff
@@ -173,15 +180,24 @@ in
...
 
+    warnings = [
+      (lib.optionalString nvidiaDockerActive
+        "You have set virtualisation.docker.enableNvidia. This option is deprecated, please set hardware.nvidia-jetpack.container-toolkit.enable instead."
+      )
+      (lib.optionalString nvidiaPodmanActive
+        "You have set virtualisation.podman.enableNvidia. This option is deprecated, please set hardware.nvidia-jetpack.container-toolkit.enable instead."
+      )
+    ];
+
```
jmbaur: Should use `lib.optionals` instead of `lib.optionalString`. The way it is currently, there could exist two warnings with empty strings, which might get filtered out (?), but probably better to not have at all

modules/default.nix
```diff
@@ -374,7 +390,7 @@ in
     ];
 
     systemd.services.nvidia-cdi-generate = {
```
jmbaur:
```suggestion
    systemd.services.nvidia-cdi-generate = mkIf cfg.container-toolkit.enable {
```

I would prefer not creating the systemd unit at all if our container-toolkit option is not enabled. The behavior of nixpkgs when `systemd.services.<name>.enable` is false is to create a masked unit, which means the unit still exists on the running system. I think not having the unit exist at all is better. `systemd.services.<name>.enable` defaults to `true`, so we can omit setting it at all.

PR #312, opened 2025-07-01
JetPack 6 Display Kernel Driver Fixes

elliotberman:
###### Description of changes

Display support for JetPack 6 is still in progress.

###### Testing

- [x] Display driver probes on orin-agx-devkit with JP6

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -288,14 +288,14 @@ in
 
       boot.kernelModules =
         [ "nvgpu" ]
-        ++ lib.optionals cfg.modesetting.enable [
-          "tegra-udrm" # For Xavier`
-          "nvidia-drm" # For Orin
-        ];
+        ++ lib.optionals (cfg.modesetting.enable && checkValidSoms [ "xavier" ]) [ "tegra-udrm" ]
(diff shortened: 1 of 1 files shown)


modules/default.nix
```diff
@@ -288,14 +288,14 @@ in
 
       boot.kernelModules =
         [ "nvgpu" ]
-        ++ lib.optionals cfg.modesetting.enable [
-          "tegra-udrm" # For Xavier`
-          "nvidia-drm" # For Orin
-        ];
+        ++ lib.optionals (cfg.modesetting.enable && checkValidSoms ["xavier"]) [ "tegra-udrm" ]
+        ++ lib.optionals (cfg.modesetting.enable && checkValidSoms ["orin"]) [ "nvidia-drm" ];
```
danielfullmer: Does this work if som == "generic" ? In that case if modesetting.enable = true we still want to enable both kernel modules.

danielfullmer (changes requested):

modules/default.nix
elliotberman: `checkValidSoms` does that: https://github.com/anduril/jetpack-nixos/blob/69848c37698fa01430b95938f41b8ae68656ac88/modules/default.nix#L33

PR #317, opened 2025-07-03
CDI: wait until required devices exist

jmbaur:
###### Description of changes

In my experience, it takes about two seconds for the probe of `nvgpu` to finish.

```
[    5.812365] nvgpu: 17000000.ga10b          nvgpu_nvhost_syncpt_init:135  [INFO]  syncpt_unit_base 60000000 syncpt_unit_size 4000000 size 10000
[   36.047096] nvgpu: 17000000.ga10b                  gk20a_scale_init:541  [INFO]  enabled scaling for GPU
[   36.049086] nvgpu: 17000000.ga10b             railgate_enable_store:323  [INFO]  railgate is disabled.
[   36.071030] nvgpu: 17000000.ga10b             railgate_enable_store:323  [INFO]  railgate is enabled.
[   36.071093] nvgpu: 17000000.ga10b             railgate_enable_store:323  [INFO]  railgate is disabled.
```

This can cause a race condition where nvidia-cdi-generate.service runs before we have access to all the character devices we want in our CDI config. Thus, we should explicitly wait until all character devices exist prior to continuing with our CDI generation. The device `/dev/nvgpu/igpu0/power` is available pretty much immediately, so we wait until one of the other devices is available prior to doing the thing. The `/dev/nvgpu/igpu0/ctrl` device sounds the most generic, so let's use that one.

###### Testing

Hard to reproduce the race condition, however in my testing, this at least does not break the CDI generation.


modules/default.nix
```diff
@@ -442,11 +442,21 @@ in
...
+        ];
         script =
           let
             exe = lib.getExe pkgs.nvidia-jetpack.nvidia-ctk;
           in
           ''
+            # /dev/nvgpu/igpu0/power is available as soon as the nvgpu kernel
+            # module is loaded, however we want to wait until further character
+            # devices exist before generating the CDI yaml.
+            udevadm wait --settle /dev/nvgpu/igpu0/ctrl
```
elliotberman: There are a lot of other devices mentioned in the csv file. Why was nvgpu picked over others? Could we use some fancy shell scripting to wait for everything to exist (with a timeout)? Does the cdi generator checks if the files mentioned in CSV exist and remove them from the CDI yaml if they don't?

jmbaur: The entire set of devices listed in the CSV file shipped with the BSP is:


```
/dev/fb0 /dev/fb1 /dev/nvhost-as-gpu /dev/nvhost-ctrl /dev/nvhost-ctrl-gpu /dev/nvhost-dbg-gpu /dev/nvhost-gpu /dev/nvhost-nvdec /dev/nvhost-nvdec1 /dev/nvhost-prof-gpu /dev/nvhost-vic /dev/nvhost-ctrl-nvdla0 /dev/nvhost-ctrl-nvdla1 /dev/nvhost-nvdla0 /dev/nvhost-nvdla1 /dev/nvidiactl /dev/nvmap /dev/tegra_dc_0 /dev/tegra_dc_1 /dev/tegra_dc_ctrl /dev/nvhost-msenc /dev/nvhost-nvenc1 /dev/nvhost-nvjpg /dev/nvhost-nvjpg1 /dev/nvhost-ctrl-pva0 /dev/nvhost-ctrl-pva1 /dev/nvhost-ofa /dev/nvhost-pva0 /dev/nvhost-pva1 /dev/nvgpu/igpu0/as /dev/nvgpu/igpu0/channel /dev/nvgpu/igpu0/ctrl /dev/nvgpu/igpu0/ctxsw /dev/nvgpu/igpu0/dbg /dev/nvgpu/igpu0/power /dev/nvgpu/igpu0/prof /dev/nvgpu/igpu0/prof-ctx /dev/nvgpu/igpu0/prof-dev /dev/nvgpu/igpu0/tsg /dev/host1x
```

I'll change the udevadm call to wait for all devices, though I think we should still keep the dependency on `modprobe@nvgpu.service`.

elliotberman (approved):

modules/default.nix
```diff
@@ -442,18 +442,30 @@ in
...
         script =
           let
             exe = lib.getExe pkgs.nvidia-jetpack.nvidia-ctk;
           in
           ''
+            # Wait until all devices are present before generating CDI
+            # configuration. Also ensure that we aren't passing any directories
+            # or glob patterns to udevadm (Jetpack 6 CSVs seem to add these,
+            # though the Jetpack 5 CSVs do not have them).
+            udevadm wait --settle $(find ${pkgs.nvidia-jetpack.l4tCsv}/ -type f -exec grep '/dev/' {} \; | grep -v -e '\*' -e 'by-path' | cut -d',' -f2 | tr -d '\n')
```
elliotberman: I stumbled on a device that reliably reproduces this issue. I pulled in this patch and found `/dev/fb0` and other devices might not exist if the display drivers aren't enabled. The `udevadm wait` is a good fix.

```suggestion
            udevadm wait --settle --timeout 10 $(find ${pkgs.nvidia-jetpack.l4tCsv}/ -type f -exec grep '/dev/' {} \; | grep -v -e '\*' -e 'by-path' | cut -d',' -f2 | tr -d '\n') || true
```

I'm not attached to the 10 seconds timeout. Maybe we can make it user configurable?

elliotberman (changes requested):

modules/default.nix
jmbaur: The default timeout is 120 seconds, any reason we don't want to allow for that much time to pass in the hopes of producing a fully valid CDI configuration? In most cases, the full timeout length shouldn't be reached anyways, limiting this to 10 seconds doesn't feel right, though I'm open to being persuaded.

elliotberman: I must've been impatient and not waited two minutes. I can imagine someone might have performance indicators measuring bootup to container running which need to be shorter than 2 minutes.

jmbaur: I don't think two mins is a short amount of time at all, I would definitely be impatient as well, I guess I'm more surprised the timeout was reached at all. Do you know which char device failed to show up?

elliotberman: /dev/fb0 was the first on the list and didn't show up. I think it makes sense because display drivers weren't enabled. I immediately went to implement/test a timeout and didn't see what else was missing. I can check on this device again if we'd like a better list

jmbaur: I'm OK putting in a shorter timeout, I guess I'm just more concerned that there was a missing device at all, possibly indicating a problem that should be looked into further.

jmbaur: I think it makes sense to say if the device doesn't show up in 10 or so seconds, there's no reason we should believe it will show up in 120 seconds. In general, I like to avoid adding options unless absolutely or obviously needed, so I say we start with a hard-coded timeout and if we or others run into issues, an option can be added. I'm assuming in your case, /dev/fb0 was not available, but most other char devices were, so the CDI config wasn't completely incomplete?

modules/default.nix
```diff
@@ -442,18 +442,30 @@ in
...
         script =
           let
             exe = lib.getExe pkgs.nvidia-jetpack.nvidia-ctk;
           in
           ''
+            # Wait until all devices are present before generating CDI
+            # configuration. Also ensure that we aren't passing any directories
+            # or glob patterns to udevadm (Jetpack 6 CSVs seem to add these,
+            # though the Jetpack 5 CSVs do not have them).
+            udevadm wait --settle --timeout 10 $(find ${pkgs.nvidia-jetpack.l4tCsv}/ -type f -exec grep '/dev/' {} \; | grep -v -e '\*' -e 'by-path' | cut -d',' -f2 | tr -d '\n')
```
elliotberman: Needs `|| true` so `udevadm` doesn't kill the service if it times out.

```suggestion
            udevadm wait --settle --timeout 10 $(find ${pkgs.nvidia-jetpack.l4tCsv}/ -type f -exec grep '/dev/' {} \; | grep -v -e '\*' -e 'by-path' | cut -d',' -f2 | tr -d '\n') || true
```

elliotberman (changes requested):

modules/default.nix
jmbaur: ah yep, thank you!

PR #323, opened 2025-07-18
Introduce jetsonCheckSpecCompatability pre-switch check

elliotberman:
###### Description of changes

We can check whether the bup for a NixOS configuration matches the machine we're trying to deploy the NixOS system to. This aims to improve user experience: if the firmware doesn't support the expected board, the capsule update will fail and it's not clear why unless you inspect the serial logs.

The check can be manually skipped by setting the
SKIP_TEGRA_COMPAT_CHECK environment variable.

###### Testing

 - [x] `switch-to-configuration` with a matching system still works
 - [x] `switch-to-configuration` with a non-matching system fails:
```
This machine is not compatible with the platform specs for this NixOS system!
This machine is a: jetson-agx-orin-devkit
This NixOS system supports:  jetson-orin-nano-devkit
Pre-switch check 'jetsonCheckSpecCompatability' failed
Pre-switch checks failed
```


modules/flash-script.nix
```diff
@@ -436,6 +436,16 @@ in
...
     };
 
+    system.preSwitchChecks = lib.mkIf canUpdateFirmware {
+      jetsonCheckSpecCompatability = "${lib.getExe (pkgs.writeShellApplication {
+      name = "ota-check-tegra-compatability-spec";
+      text = ''
+        [ -n "''${SKIP_TEGRA_COMPAT_CHECK+set}" ] && exit 0
+        exec ${lib.getExe' pkgs.nvidia-jetpack.otaUtils "ota-check-compat"} ${pkgs.nvidia-jetpack.bupSpecs}
+      '';
+    })}";
```
ConnorBaker: Double-check that this was formatted, it looks odd to me

pkgs/ota-utils/ota-check-compat.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# If the tokens do not match, return 1 indicating no match
+if [[ "$token1" != "$token2" ]]; then
+return 1
+fi
+done
+
+# All tokens match (ignoring empty tokens), return 0 for success
+return 0
+}
```
ConnorBaker: Consider running through `shfmt`

elliotberman:
Superseded by #369

PR #326, opened 2025-07-22
modules: newer JetPack can use packages from older JetPack

ConnorBaker:
###### Description of changes

NVIDIA generally provides the ability to run older software releases on newer firmware/drivers. As a result, we allow newer versions of JetPack to reference packages from previous releases, but forbid older versions from referencing packages from newer releases.

###### Testing

🤷‍♂️

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -10,7 +10,9 @@ let
     mkEnableOption
     mkIf
     mkOption
+    nameValuePair
     types
+    versionOlder
     ;
 
   cfg = config.hardware.nvidia-jetpack;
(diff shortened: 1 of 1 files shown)


danielfullmer:
> NVIDIA generally provides the ability to run older software releases on newer firmware/drivers
Is this true for Jetson? I always thought this was YMMV.

ConnorBaker:
Their docs on forward compatibility (https://docs.nvidia.com/cuda/cuda-for-tegra-appnote/#deployment-considerations-for-cuda-upgrade-package) references binary compatibility (https://docs.nvidia.com/cuda/cuda-c-best-practices-guide/index.html#binary-compatibility) in their support matrices, making me believe that this is the case.

However, the giant disclaimer of liability at the bottom of every page does indicate YMMV :)

modules/default.nix
```diff
@@ -227,14 +229,25 @@ in
         (
           let
             otherJetpacks = builtins.filter (v: v != cfg.majorVersion) jetpackVersions;
```
elliotberman, line 231: Move `otherJetpacks` so we can delete this `let ... in`?

elliotberman (commented):
> NVIDIA generally provides the ability to run older software releases on newer firmware/drivers.

Could we limit the backwards compat to only CUDA packages? I don't think other L4T packages will generally be backwards compatible.

Could expand the `combined-test.nix` to also run the older CUDA tests?

PR #331, opened 2025-08-06
Use nvidia-container-toolkit from Nixpkgs

ConnorBaker:
###### Description of changes

Removes the vendored expression for `nvidia-container-toolkit` and switches to using the one provided by upstream Nixpkgs.

The flake lockfile bump puts us on a new enough version of 25.05 that `nvidia-container-toolkit` has version 1.17.8, which has patched CVE-2025-23266 and CVE-2025-23267 (https://nvidia.custhelp.com/app/answers/detail/a_id/5659).

###### Testing

To be run in CI.


ConnorBaker:
@elliotberman would you review when you get a chance?

ConnorBaker:
@jmbaur you may also be interested in this, if only for nostalgia :)

pkgs/containers/l4t-csv.nix
```diff
@@ -1,5 +1,62 @@
...
+
+  pname = "l4t-csv";
+  version = l4tMajorMinorPatchVersion;
+
+  src = "${bspSrc}/nv_tegra/config.tbz2";
+
+  sourceRoot = "etc/nvidia-container-runtime/host-files-for-container.d";
+
+  # We keep track of the file names so we can use them in the module system to enable nvidia-container-toolkit.
+  # Also allows us to make sure we're copying over everything we should.
```
elliotberman, line 19: What happens if there are additional files added in a minor JetPack release? I'm concerned we would silently miss them until someone reports something not functioning.

ConnorBaker: That's why in the `installPhase`, we ensure that all the files we currently track exist and are moved over, and in the `installCheckPhase` we ensure that there are no files left (and therefor our list is complete).

elliotberman (approved):
Not a fan that switching to an upstream implementation somehow increased our LOC but that's not the point

danielfullmer:
Tested working on Orin AGX with Jetpack 6 and a recent version of nixpkgs

PR #334, opened 2025-08-07
Report progress of the initrd flashing script

elliotberman:
###### Description of changes

The initrd flash script doesn't report progress back to the flashing
station (host PC). When a carrier board doesn't have a serial port, this
is unfortunate because the user doesn't have any way of knowing whether
the flashing is finished, failed, or succeeded. Fortunately, there's a
USB port we can expose an ACM gadget through and pipe the script's
progress through.

###### Testing

- [x] flash-orin-agx-devkit
- [x] flash-orin-agx-devkit-jp5
- [ ] Xavier flashing
- [ ] All carrier boards initrd flashing


device-pkgs/default.nix
```diff
@@ -118,31 +120,127 @@ let
...
-        echo "You may watch the progress of this on the device's serial port"
-      '';
+      runtimeInputs = [ picocom ];
+      text =
+        let
+          inherit (flashInitrd.passthru) manufacturer product serialnumber;
+          serialPortId = "usb-${manufacturer}_${product}_${serialnumber}-if00";
+        in
+        ''
+          ${mkRcmBootScript {
```
ConnorBaker: Please consider floating this out and using string concatenation rather than templating large expressions

device-pkgs/default.nix
```diff
@@ -118,31 +120,127 @@ let
...
+            # persisted to the firmware storage medium for future booting. We
+            # don't want to influence the boot order of the temporary edk2 since
+            # this may cause that edk2 to boot from something other than our
+            # intended flashing kernel/initrd combo (e.g. disk or network). Since
+            # the edk2 build that we actually persist to the board is embedded in
+            # the initrd used for flashing, we have the desired boot order (as
+            # configured in nix) in there and is not affected dynamically during
+            # the flashing procedure. NVIDIA ensures that when the device is
+            # using RCM boot, only the boot mode named "boot.img" is used (see https://gist.github.com/jmbaur/1ca79436e69eadc0a38ec0b43b16cb2f#file-flash-sh-L1675).
+            additionalDtbOverlays = (lib.filter (path: (path.name or "") != "DefaultBootOrder.dtbo") cfg.flashScriptOverrides.additionalDtbOverlays) ++ [(pkgs.deviceTree.compileDTS {
```
ConnorBaker:
```suggestion
            additionalDtbOverlays = lib.filter (path: (path.name or "") != "DefaultBootOrder.dtbo") cfg.flashScriptOverrides.additionalDtbOverlays ++ [(pkgs.deviceTree.compileDTS {
```

unnecessary parenthesis

overlay-with-config.nix
```diff
@@ -78,25 +78,66 @@ final: prev: (
...
 
             for mod in ${builtins.toString modules}; do
               modprobe -v $mod
             done
 
+            mount -t configfs none /sys/kernel/config
+            if [ -e /sys/kernel/config/usb_gadget ] ; then
+              # https://origin.kernel.org/doc/html/v5.10/usb/gadget_configfs.html
+              gadget=/sys/kernel/config/usb_gadget/g.1
+              mkdir $gadget
```
ConnorBaker, line 111: No need to use `mkdir -p` in case a parent doesn't exist?

ConnorBaker (commented):
A few comments/questions.

overlay-with-config.nix
elliotberman: I check that `/sys/kernel/config/usb_gadget` exists and these directories are created by the kernel

pkgs/flash-from-device/flash-from-device.sh
```diff
@@ -4,12 +4,20 @@
 # Licensed under MIT.
 
 set -euo pipefail
+set -x
```
danielfullmer: remove?

device-pkgs/initrdflash-script.nix
```diff
@@ -0,0 +1,141 @@
...
+  echo
+
+  if [ ! -e /dev/serial/by-id/${serialPortId} ] ; then
+    echo "Jetson device not detected, please monitor the serial port or wait 5-15 minutes"
+    exit 2
+  fi
+
+  logs=$(mktemp)
+  echo "^a^x to kill the terminal"
+  picocom --quiet --logfile "$logs" /dev/serial/by-id/${serialPortId} || true
```
danielfullmer: Could we use "expect" or something like that here instead of picocom?

device-pkgs/initrdflash-script.nix
```diff
@@ -0,0 +1,141 @@
...
+    ((++counter))
+  done
+  echo
+
+  if [ ! -e /dev/serial/by-id/${serialPortId} ] ; then
+    echo "Jetson device not detected, please monitor the serial port or wait 5-15 minutes"
+    exit 2
+  fi
+
+  logs=$(mktemp)
```
danielfullmer: I'd prefer if we didn't have to have a temporary log file here, and either use a pipe or "expect" instead.

device-pkgs/initrdflash-script.nix
```diff
@@ -0,0 +1,141 @@
...
+  echo "^a^x to kill the terminal"
+  picocom --quiet --logfile "$logs" /dev/serial/by-id/${serialPortId} || true
+
+  if ! grep -q "Flashing platform firmware successful" "$logs" ; then
+    rm "$logs"
+    echo "Flashing may have failed"
+    exit 1
+  fi
+  rm "$logs"
```
danielfullmer: If we're going to have a temporary log file, you'd need to `rm` it in a bash trap.

modules/devices.nix
```diff
@@ -146,20 +146,6 @@ in
           partitionTemplate = mkDefault (filterPartitions defaultPartitionsToRemove "${pkgs.nvidia-jetpack.bspSrc}/bootloader/${partitionTemplateDirectory}/cfg/flash_l4t_t194_spi_emmc_p3668.xml");
         })
       ];
-
-    boot.kernelPatches = lib.mkIf (cfg.som == "orin-nx" && lib.versions.majorMinor config.system.build.kernel.version == "5.10") [
-      {
-        name = "disable-usb-otg";
-        patch = null;
-        # TODO: Having these options enabled on the Orin NX currently causes a
```
danielfullmer, line 154: Is this no longer the case?

overlay-with-config.nix
```diff
@@ -79,25 +79,66 @@ final: prev: (
...
-          modules = if lib.versions.majorMinor config.system.build.kernel.version == "5.10" then [ "qspi_mtd" "spi_tegra210_qspi" "at24" "spi_nor" ] else [ "mtdblock" "spi_tegra210_quad" ];
+          spiModules = if lib.versions.majorMinor config.system.build.kernel.version == "5.10" then [ "qspi_mtd" "spi_tegra210_qspi" "at24" "spi_nor" ] else [ "mtdblock" "spi_tegra210_quad" ];
+          usbModules = if lib.versions.majorMinor config.system.build.kernel.version == "5.10" then [ ] else [ "libcomposite" "udc-core" "tegra-xudc" "xhci-tegra" "u_serial" "usb_f_acm" ];
+          modules = spiModules ++ usbModules ++ cfg.flashScriptOverrides.additionalInitrdFlashModules;
           modulesClosure = prev.makeModulesClosure {
             rootModules = modules;
             kernel = config.system.modulesTree;
             firmware = config.hardware.firmware;
-            allowMissing = false;
+            allowMissing = true;
```
danielfullmer: Do we need this? If so, could you add a comment saying why?

overlay-with-config.nix
```diff
@@ -79,25 +79,66 @@ final: prev: (
...
+              ln -s $gadget/functions/acm.usb0 $gadget/configs/c.1/
+
+              echo "$(ls /sys/class/udc | head -n 1)" >$gadget/UDC
+
+              # force into device mode if OTG and something is up with automatic detection
+              if [ -w /sys/class/usb_role/usb2-0-role-switch/role ] ; then
+                echo device > /sys/class/usb_role/usb2-0-role-switch/role
+              fi
+
+              sleep 5  # The configuration doesn't happen synchronously and takes >1 sec. 5 seconds seems like a good buffer and also gives time for host to connect
```
danielfullmer, line 133: Do we have any way to wait on this instead of just sleeping?

modules/devices.nix
elliotberman: Yes, it's no longer the case

overlay-with-config.nix
elliotberman: We could do some sort of handshake with the host and use something like `expect`. It adds complexity and we're trying to keep the ramdisk small.

danielfullmer (approved):

overlay-with-config.nix
danielfullmer: While i'd really like this--it can wait for a follow up PR

PR #337, opened 2025-08-15
Enable treefmt

elliotberman:
###### Description of changes

Enable formatter for the few shell and python scripts we have.


flake.nix
```diff
@@ -128,7 +133,15 @@
...
-      formatter = forAllSystems ({ pkgs, ... }: pkgs.nixpkgs-fmt);
+      formatter = forAllSystems ({ pkgs, ... }: (treefmt-nix.lib.evalModule pkgs (
+        { ... }:
+        {
+          projectRootFile = "flake.nix";
+          programs.nixpkgs-fmt.enable = true;
+          programs.shfmt.enable = true;
+          programs.black.enable = true;
+        }
+      )).config.build.wrapper);
```
ConnorBaker:
```suggestion
      formatter = forAllSystems ({ pkgs, ... }: (treefmt-nix.lib.evalModule pkgs {
        projectRootFile = "flake.nix";
        programs.nixpkgs-fmt.enable = true;
        programs.shfmt.enable = true;
        programs.black.enable = true;
      }).config.build.wrapper);
```

Raw attribute sets are also valid modules :)

NOTE: Haven't tested eval or run the formatter with this change, but it should work.

ConnorBaker (approved):
Very nice! All for adding formatting :)

flake.nix
```diff
@@ -1,9 +1,14 @@
 {
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs/nixos-25.05";
+
+    treefmt-nix = {
+      url = "github:numtide/treefmt-nix";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
```
bryanhonof: suggestion: You might get away with not adding a new flake input

The package in upstream nixpkgs provides a[ `passthru` that allows you to wrap the package with some config](https://github.com/NixOS/nixpkgs/blob/a58390ab6f1aa810eb8e0f0fc74230e7cc06de03/pkgs/by-name/tr/treefmt/package.nix#L33-L48). I do think you might have to define some of the formatters manually, but I do think it's worth considering to not add another input.

bryanhonof (approved):

flake.nix
elliotberman: Neat!

danielfullmer (approved):

PR #342, opened 2025-08-21
cudaPackages: align with upstream

ConnorBaker:
###### Description of changes

Aligns with upstream's changes in https://github.com/NixOS/nixpkgs/pull/406568.

`nvidia-jetpack.cudaPackages` now has a `pkgs` attribute which functions in the same way as upstream's `cudaPackages.pkgs` attribute: https://nixos.org/manual/nixpkgs/stable/#cuda-using-cudapackages-pkgs.

The default version of `nvidia-jetpack` is now determined by the default version of the CUDA package set (`cudaPackages`). Changing the default version of the CUDA package set (either through something like `cudaPackages_12.pkgs` or an overlay) now changes the default version of `nvidia-jetpack`.

Added assertions to the NixOS configuration which, when CUDA support is requested, check that the default CUDA package set can be used with the current version of JetPack NixOS. As an example, taking a JetPack 6 closure and purposefully changing the default CUDA version by applying an overlay like

```nix
final: _: { cudaPackages = final.cudaPackages_12_2; }
```

would provide this evaluation error:

```console
error:
       … while calling the 'head' builtin
         at /nix/store/5n6k4jld2n6k0cs98h9xgh35xsa03jk3-source/lib/attrsets.nix:1534:13:
         1533|           if length values == 1 || pred here (elemAt values 1) (head values) then
         1534|             head values
             |             ^
         1535|           else

       … while evaluating the attribute 'value'
         at /nix/store/5n6k4jld2n6k0cs98h9xgh35xsa03jk3-source/lib/modules.nix:1083:7:
         1082|     // {
         1083|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |       ^
         1084|       inherit (res.defsFinal') highestPrio;

       … while evaluating the option `system.build.toplevel':

       … while evaluating definitions from `/nix/store/5n6k4jld2n6k0cs98h9xgh35xsa03jk3-source/nixos/modules/system/activation/top-level.nix':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error:
       Failed assertions:
       - JetPack NixOS 6 supports CUDA 12.4 (natively) - 12.9 (with `cuda_compat`): `pkgs.cudaPackages` has version 12.2.
```

> [!IMPORTANT]
>
> If NVIDIA provides backward-compatibility guarantees for Tegra/Jetson, using an older version of CUDA packages with a newer system should be okay, but I've nested tested or looked into what's necessary to make that happen.

> [!NOTE]
>
> In order to get the NixOS assertions working for all of the different CUDA overlays I could think of, I did have to change the NixOS module such that it only wraps unsupported `nvidia-jetpack` package sets in a warning, rather than setting them to the empty attribute set. I can work around this if desired by instead turning it into a singleton which just contains `cudaPackages.cudaMajorMinorVersion`.

Added a new check, `jetpackSelectionDependsOnCudaVersion`, which uses assertions to verify the version-changing behavior made possible by `cudaPackages.pkgs` and the like function as expected.

###### Testing

- [x] CI (internal)
- [x] Manual testing in HITL (internal; in progress)
- [x] Manual verification

### DIFF
--- a/check-jetpack-selection.nix
+++ b/check-jetpack-selection.nix
@@ -1,0 +1,122 @@
+{ lib, nixpkgs, overlay, pkgs, system }:
+(
+  let
+    inherit (lib) assertMsg getAttrFromPath showAttrPath;
+    inherit (lib.versions) major;
+
+    mkPkgs = overlay': (import nixpkgs {
+      inherit system;
+      config = {
+        allowUnfree = true;
+        cudaCapabilities = [ "7.2" "8.7" ];
+        cudaSupport = true;
+      };
+      overlays = [ overlay overlay' ];
+    });
+
+    variants = {
+      pkgsCudaDefault = mkPkgs (_: _: { });
+
+      pkgsCuda11 = mkPkgs (final: _: {
+        cudaPackages = final.cudaPackages_11;
+      });
+
+      pkgsCuda12 = mkPkgs (final: _: {
+        cudaPackages = final.cudaPackages_12;
+      });
+    };
+
+    mkAsserts =
+      { pkgsPath
+        # ^ Path to the root of the attribute set to use as `pkgs`, through `variants`.
+        #   For example, `["variants" "pkgsCuda11" "cudaPackages_12" "pkgs"]` would use
+        #   `variants.pkgsCuda11.cudaPackages_12.pkgs` as the root package set.
+      , expectedX86CudaMajorMinorVersion
+      , expectedAarch64CudaMajorMinorVersion
+      , expectedJetPackMajorVersion
+      , expectedJetPackCudaMajorMinorVersion
+      ,
+      }:
+      let
+        prefix = showAttrPath pkgsPath;
+        inherit (getAttrFromPath pkgsPath variants) cudaPackages nvidia-jetpack;
+        actualCudaMajorMinorVersion = cudaPackages.cudaMajorMinorVersion;
+        actualJetPackMajorVersion = major nvidia-jetpack.jetpackMajorMinorPatchVersion;
+        actualJetPackCudaCudaMajorMinorVersion = nvidia-jetpack.cudaPackages.cudaMajorMinorVersion;
+      in
+      assert assertMsg (system == "x86_64-linux" -> actualCudaMajorMinorVersion == expectedX86CudaMajorMinorVersion)
+        "${prefix}: Expected CUDA ${expectedX86CudaMajorMinorVersion} to be the default for x86_64-linux, not CUDA ${actualCudaMajorMinorVersion}";
+
+      assert assertMsg (system == "aarch64-linux" -> actualCudaMajorMinorVersion == expectedAarch64CudaMajorMinorVersion)
+        "${prefix}: Expected CUDA ${expectedAarch64CudaMajorMinorVersion} to be the default for aarch64-linux, not CUDA ${actualCudaMajorMinorVersion}";
+
+      assert assertMsg (actualJetPackMajorVersion == expectedJetPackMajorVersion)
+        "${prefix}: Expected JetPack ${expectedJetPackMajorVersion} to be the default, not JetPack ${actualJetPackMajorVersion}";
+
+      assert assertMsg (actualJetPackCudaCudaMajorMinorVersion == expectedJetPackCudaMajorMinorVersion)
+        (
+          "${prefix}: Expected JetPack CUDA ${expectedJetPackCudaMajorMinorVersion} to be the default,"
+          + " not JetPack CUDA ${actualJetPackCudaCudaMajorMinorVersion}"
+        );
+
+      # On aarch64-linux, the global CUDA package set should be *our* CUDA package set
+      assert assertMsg (system == "aarch64-linux" -> expectedJetPackCudaMajorMinorVersion == actualCudaMajorMinorVersion)
+        "${prefix}: Expected JetPack CUDA ${expectedJetPackCudaMajorMinorVersion} to be the default CUDA for aarch64-linux, not CUDA ${actualCudaMajorMinorVersion}";
+
+      # We have to return something so give the empty attribute set.
+      { };
+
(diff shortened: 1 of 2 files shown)


mk-overlay.nix
```diff
@@ -103,28 +107,131 @@ makeScope final.newScope (self: {
...
+          final.extend (
+            final: _: {
+              recurseForDerivations = false;
+              # The CUDA package set will be available as cudaPackages_x_y, so we need only update the aliases for the
+              # minor-versioned and unversioned package sets.
+              # cudaPackages_x = cudaPackages_x_y
+              ${cudaPackagesMajorVersionName} = final.${cudaPackagesMajorMinorVersionName};
+              # cudaPackages = cudaPackages_x
+              ${cudaPackagesUnversionedName} = final.${cudaPackagesMajorVersionName};
+              # Change the default nvidia-jetpack as well.
```
elliotberman:
```suggestion
              # nvidia-jetpack = final.nvidia-jetpackX
```

overlay.nix
```diff
@@ -32,5 +57,60 @@ final: prev:
...
+      final.nvidia-jetpack5
+
+    # CUDA 12.4 - 12.9 is supported either natively (12.4) or through cuda_compat (everything else)
+    else if cudaOlder "13.0" then
+      final.nvidia-jetpack6
+
+    # CUDA 13.0+ is supported by JetPack 7, which we don't yet package.
+    # Use JetPack 6 to avoid throwing.
+    else
+      warn "cudaPackages before 12.9 are not supported (observed version ${cudaMajorMinorVersion})" final.nvidia-jetpack6;
```
elliotberman: Do we need to warn? The default was previously jetpack 5, shall we keep it that way?

overlay.nix
```diff
@@ -32,5 +57,60 @@ final: prev:
...
+
+  # Set cudaPackage package sets to our JetPack-constructed package sets if we are targeting Jetson capabilities.
+  # NOTE: We cannot lift the conditionals out further without causing infinite recursion, as the fixed-point would be
+  # used to determine the presence/absence of attributes.
+  cudaPackages_11_4 =
+    assert final.nvidia-jetpack5.cudaPackages.cudaMajorMinorVersion == "11.4";
+    if redistSystem == "linux-aarch64" then
+      final.nvidia-jetpack5.cudaPackages
```
elliotberman, line 92: Does this cause issues if:
1. cudaCapabilities has a mix of both jetson and non-jetson GPUs
2. SBSA machine wants to use CUDA 11.4 (or any other overridden version below)

If yes, I think that could be reasonable and we just need to acknowledge it.

elliotberman (commented):
I think I've understood the big picture here, left a couple minor comments. Will take another pass at it soon once it's had time to digest.

---

> Aligns with upstream's changes in https://github.com/NixOS/nixpkgs/pull/406568.

Is upstream a dependency?

> The default version of nvidia-jetpack is now determined by the default version of the CUDA package set (cudaPackages).

Can you justify why? We talked about this offline, but would be good to get it in writing for our future selves. Super ideally, the comment can be done in overlay.nix and in the PR description.

mk-overlay.nix
ConnorBaker: Fixed in latest (pending push)

overlay.nix
ConnorBaker: Removed the warning in the latest (pending push).

I think the logic should remain (older than 12.3 uses JP5, newer than that uses JP6 -- though that's a bit problematic because 12.3 isn't supported by either, IIRC). The version of `cudaPackages` drives the version of `nvidia-jetpack`; the default is still `nvidia-jetpack5` because our overlay sets the default `cudaPackages` to `cudaPackages_11`.

overlay.nix
ConnorBaker: 1. `cudaCapabilities` cannot have a mix of both Jetson and non-Jetson GPUs (pre-Thor, anyway) since there'd be ambiguity as to which redistributables to use. When `stdenv.hostPlatform.system` is `aarch64-linux`, `redistSystem` is `linux-sbsa` if there are no Jetson CUDA capabilities and `linux-aarch64` if there are any (which also means they are all Jetson CUDA capabilities because they cannot mix). (Things get a little more complicated with Thor, but I'm handling that upstream.)
2. We don't package debians for SBSA here, so support is limited to whatever upstream has (we always take the branch which passes through the original, unchanged); I don't think SBSA is well supported by the redistributables upstream before 11.8.

ConnorBaker:
Updated to address comments and force-pushed.

Added a comment to the version selection of `nvidia-jetpack`:

> Due to the interplay between JetPack releases and supported CUDA versions, the choice of CUDA version drives the
> version of nvidia-jetpack made the default to avoid the need to maintain tedious overlays and ensures the two stay
> in sync by default.
>
> Thanks to the functionality added in https://github.com/NixOS/nixpkgs/pull/406568, we can build packages
> using variants of Nixpkgs through the Flake CLI. See https://nixos.org/manual/nixpkgs/stable/#cuda-using-cudapackages-pkgs
> for an example.
>
> Consider a user trying to build a hypothetical package foo which works with all versions of nvidia-jetpack. If
> they build .#blarg (assuming proper configuration of Nixpkgs and use of our overlay), blarg receives
> nvidia-jetpack5 as nvidia-jetpack and nvidia-jetpack5.cudaPackages (CUDA 11.4) as cudaPackages.
> If they build .#cudaPackages_12.pkgs.blarg, blarg receives nvidia-jetpack6 as nvidia-jetpack and
> nvidia-jetpack6.cudaPackages (CUDA 12.6) as cudaPackages. If the version of nvidia-jetpack did not depend on
> the version of the CUDA package set, blarg would have received nvidia-jetpack5 as nvidia-jetpack (since it would
> stay unchanged) and nvidia-jetpack6.cudaPackages (CUDA 12.6) as cudaPackages -- this is likely unintentional!

PR #343, opened 2025-08-25
Add isp-setup service

sm1ll3r:
###### Description of changes

Creates a service that will copy all files under `nvcam` specified in `hardware.nvidia-jetpack.ispPkgs` into the `/var/nvidia/nvcam` directory. Whenever this list of pkgs gets updated the `/var/nvidia/nvcam` directory will be recreated. 

###### Testing

Verified files were copied into /var/nvidia/nvcam on an orin-agx. nvargus was able to read the .isp files and image captured using nvargus pipeline had isp configuration applied.

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -172,6 +172,16 @@ in
         description = "Enable boot.kernelParams default console configuration";
       };
 
+      ispPkgs = mkOption {
+        type = types.listOf types.package;
+        default = [ ];
+        description = ''
+          The list of packages that contain isp files. This
+          will copy any files in the /nvcam directory of each package to the
+          /var/nvidia/nvcam directory on the device at boot time.
+        '';
+      };
+
       console.args = mkOption {
         internal = true;
         type = types.listOf types.str;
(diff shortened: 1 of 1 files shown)


modules/isp-setup.nix
```diff
@@ -0,0 +1,39 @@
...
+    fi
+  '';
+in
+{
+  options = {
+    services.isp-setup = {
+      
+      enable = mkEnableOption "ISP setup";
+
+      path = mkOption {
```
danielfullmer: This should probably be an option under `hardware.jetpack-nixos` instead of something separate.
Could the option just be a list of packages which provide isp files? It'd be empty by default, so I don't see the need for an `enable` field.

danielfullmer (changes requested):
What happens if someone later removes isp files, how can we remove them from /var/nvidia/nvcam?
Do we need to clear the cache files nvidia makes here? (Could you document what those are?)

What can we do to test this functionality?

danielfullmer:
> Verified on orin-agx with nvargus.

Verified what, exactly?

sm1ll3r:
> What happens if someone later removes isp files, how can we remove them from /var/nvidia/nvcam? Do we need to clear the cache files nvidia makes here? (Could you document what those are?)
> 
> What can we do to test this functionality?

Per the [forums](https://forums.developer.nvidia.com/t/switching-isp-files-for-day-and-night-mode-streaming-w-o-restart/292463/4) (since there are no docs :/) it seems the nvcam_cache_*.bin and serial_no_*.bin speed up camera pipeline initialization and only need to be updated when the underlying camera_override isp files have changed.
Given that I was thinking about storing the hash of the list of packages given that have isp files and if that updates then recreate the /var/nvidia/nvcam dir?

For testing we would need to verify that nvargus or libargus initialize and are able to read the override.isp files at /var/nvidia/nvcam

modules/isp-setup.nix
sm1ll3r: Went ahead with this it's now a list of paths if empty it wont start the service.

modules/default.nix
```diff
@@ -169,6 +169,16 @@ in
         type = types.bool;
         description = "Enable boot.kernelParams default console configuration";
       };
+
+      ispPkgs = mkOption {
+        type = types.listOf types.package;
+        default = [ ];
+        description = ''
+          The list of packages that contain isp files. This
+          will copy any files in the /nvcam directory to the
```
elliotberman:
```suggestion
          will copy any files in the /nvcam directory of each package to the
```

sm1ll3r: updated

danielfullmer (approved):
LGTM

PR #344, opened 2025-08-26
Updating optee's compiling environment

TanelDettenborn:
nixpkgs unstable has removed gcc12Stdenv and below support. The next "oldest" is gcc13Stdenv.

###### Description of changes
Jetson 36.x; Anticipating upcomming nixpkgs update(s) and therefore switching directly to use gcc13Stdenv. Officially NVIDIA uses gcc11.

Jetson 35.x: Keeps gcc9 as per NVIDIA recommends.

###### Testing
NX target evaluates and compiles. Boots with iso_minimal.

### DIFF
--- a/mk-overlay.nix
+++ b/mk-overlay.nix
@@ -88,9 +88,15 @@ makeScope final.newScope (self: {
     edk2-jetson uefi-firmware;
 
   inherit (final.callPackages ./pkgs/optee {
-    # Nvidia's recommended toolchain is gcc9:
-    # https://nv-tegra.nvidia.com/r/gitweb?p=tegra/optee-src/nv-optee.git;a=blob;f=optee/atf_and_optee_README.txt;h=591edda3d4ec96997e054ebd21fc8326983d3464;hb=5ac2ab218ba9116f1df4a0bb5092b1f6d810e8f7#l33
-    stdenv = final.gcc9Stdenv;
+    # As today as this comment is written then nixpkgs unstabble has removed
+    # gcc12Stdenv and below support. The next "oldest" is gcc13Stdenv.
+    #
+    # Jetson 36.x; Anticipating upcomming nixpkgs updates and
+    # therefore swithching directly to use gcc13Stdenv. Officially NVIDIA
+    # uses gcc11
+    #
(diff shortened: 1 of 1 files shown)


brianmcgillion (approved):
LGTM.

mk-overlay.nix
```diff
@@ -88,9 +88,15 @@ makeScope final.newScope (self: {
...
-    stdenv = final.gcc9Stdenv;
+    # As today as this comment is written then nixpkgs unstabble has removed
+    # gcc12Stdenv and below support. The next "oldest" is gcc13Stdenv.
+    #
+    # Jetson 36.x; Anticipating upcomming nixpkgs updates and
+    # therefore swithching directly to use gcc13Stdenv. Officially NVIDIA
+    # uses gcc11
+    #
+    # Jetson 35.x: Keeping gcc9 as per NVIDIA recommends
+    stdenv = if self.l4tAtLeast "36" then final.gcc13Stdenv else final.gcc9Stdenv;
```
elliotberman, line 99: JetPack 5 devices remain broken with this change. I suspect we can upgrade all OP-TEE versions to gcc13.

danielfullmer:
@jmbaur If I remember correctly, you pinned this to GCC9 in response to some bug you found.  Do you recall what that was?

TanelDettenborn:
Default stdenv *seems* to work with Orin NX devkit (r35+iso_min_jp5 and r36+iso_min). Claim *work* need to take with a grain of salt, because I do not have tests. Devkit boots into console. There might be a subtle cases where it might matter if it is Bootlin GCC 9 or 11 (as NVIDIA recommends). 

I would agree with default stdenv approach, because then it would be consistent across jetpack.

If/when agreed then I can update PR.


Not directly related to this PR, but relates with tests and OP-TEE: #[149](https://github.com/anduril/jetpack-nixos/pull/149). Jetson BSP has OP-TEE's tests and in this case running those would be better than nothing. I recon that my PR might have been bulky and therefore I can make something minimalist..

danielfullmer:
Resolved in: https://github.com/anduril/jetpack-nixos/pull/360

Thanks for pointing this out and for your PR!

PR #345, opened 2025-08-26
[WIP] Jetpack 7 and Thor AGX

elliotberman:
###### Description of changes

Adding JetPack 7, slowly.

###### Testing

- Building only for now


overlay.nix
```diff
@@ -32,5 +32,20 @@ final: prev:
     final
     prev;
 
+  nvidia-jetpack7 = import ./mk-overlay.nix
+    {
+      jetpackMajorMinorPatchVersion = "7.0";
+      l4tMajorMinorPatchVersion = "38.2.0";
+      cudaMajorMinorPatchVersion = "13.0.0";
```
bryanhonof: question: is jetpack-nixos re-packaging cudaPackages itself?

When I was trying to do this bump myself yesterday, it wasn't quite clear to me the relationship between upstream cudaPackages, and the cudaPackages in this repo. Is this something we'll have to wait for upstream nixpkgs to bump, or is jetpack-nixos actually re-packaging all of cudaPackages itself?

overlay-with-config.nix
```diff
@@ -25,12 +25,14 @@ final: prev: (
...
         else if lib.hasPrefix "orin-" cfg.som then "t234"
         else if lib.hasPrefix "xavier-" cfg.som then "t194"
+        else if lib.hasPrefix "thor-" cfg.som then "t264"
         else throw "Unknown SoC type";
 
       chipId =
         if cfg.som == null then null
         else if lib.hasPrefix "orin-" cfg.som then "0x23"
         else if lib.hasPrefix "xavier-" cfg.som then "0x19"
+        else if lib.hasPrefix "thor-" cfg.som then "0x26"
```
bryanhonof: question: where can I find these magic numbers myself?

Out of curiosity, I wasn't able to find these in NVIDIA's docs. But there's a high chance I was looking in the wrong places. 😅

overlay.nix
elliotberman: We're repackaging CUDA for JetPack 5 and 6 because it's a separate version of CUDA. JetPack 7 claims to use the regular upstream CUDA 13 implementation. So, I'm unsure what we're doing for JetPack 7 yet, haven't gotten that far.

overlay-with-config.nix
elliotberman: > question: where can I find these magic numbers myself?

They're in the BSP scripts

modules/devices.nix
```diff
@@ -218,5 +210,21 @@ in
...
       ];
+    })
+    (lib.mkIf (cfg.som == "thor-agx") {
+      hardware.firmware = [
+        (extractLinuxFirmware "xusb-firmware" [ "nvidia/tegra186/xusb.bin" ])
+      ] ++ lib.optionals
+        (config.hardware.bluetooth.enable
+          || config.networking.wireless.enable
+          || config.networking.wireless.iwd.enable) [
+        "rtl_nic/rtl8153a-2.fw"
```
Princemachiavelli: This list needs to be passed to `extractLinuxFirmware`; this option expects packages not lists of string.

device-pkgs/default.nix
```diff
@@ -84,6 +87,7 @@ let
 
       flashArgs =
         [ "--rcm-boot" ]
+        ++ lib.optional (jetpackAtLeast "7") "-r"
```
danielfullmer, line 91: Add a comment about what this is and why it's needed on JP7? In particular, is there anything different about JP7 that makes it necessary, but it's not needed on JP5/6?

pkgs/r38-bsp.patch
```diff
@@ -0,0 +1,27 @@
+diff --git a/bootloader/LICENSE.bl31_t264.fip.rst b/bootloader/LICENSE.bl31_t264.fip.rst
```
danielfullmer, line 1: Is this patch file necessary at all?

elliotberman: Nix complains about broken symlinks otherwise

overlay.nix
```diff
@@ -56,6 +56,20 @@ in
...
+  nvidia-jetpack7 = import ./mk-overlay.nix
+    {
+      jetpackMajorMinorPatchVersion = "7.0";
+      l4tMajorMinorPatchVersion = "38.2.1";
+      cudaMajorMinorPatchVersion = "12.6.10"; #TODO
+
+      cudaDriverMajorMinorVersion = "580.00";
+
+      bspHash = "sha256-raHtaLeODpgHxw24e+ViturGqpXVOL9jtun4owCDcEs=";
+      bspPatches = [ ./pkgs/r38-bsp.patch ];
```
danielfullmer, line 68: See other note about whether we need this patch at all.
Also, we already have a nixos-level patch script under `hardware.nvidia-jetpack.flashScriptOverrides.patches`. Do we still need something here?

pkgs/flash-tools/default.nix
```diff
@@ -76,6 +77,7 @@ let
     '' + (lib.optionalString (!stdenv.hostPlatform.isx86) ''
       # Wrap x86 binaries in qemu
       pushd bootloader/ >/dev/null
+      # Wrap i386 binaries in qemu
```
danielfullmer, line 80: We can do it in a later PR, but it'd be nice to deduplicate the i386 / x86_64 bits below

overlay.nix
elliotberman: We have a few derivations (op-tee, kernel, containers) which rely on the bspsrc directly and not the flashScript path. We need `bspSrc` derivation to build cleanly :/

pkgs/r38-bsp.patch
danielfullmer: Can you just disable the broken symlinks checker for this derivation?

PR #348, opened 2025-08-29
Create libEGL symlink for libnvscf.so library

sm1ll3r:
###### Description of changes

Adds a service that symlinks
`/run/opengl-driver/lib/libEGL_nvidia.so.0 -> /usr/lib/aarch64-linux-gnu/tegra-egl/libEGL_nvidia.so.0`
This allows the nvargus-daemon and libargus application to function correctly.

###### Testing

- [ ] Symlink gets created
- [ ] Both the nvargus-daemon and libargus applications execute without failures.

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -408,6 +408,39 @@ in
       # Include nv_tegra_release, just so we can tell what version our NixOS machine was built from.
       environment.etc."nv_tegra_release".source = "${pkgs.nvidia-jetpack.l4t-core}/etc/nv_tegra_release";
 
+      # For some unknown reason, the libnvscf.so library has a dlopen call to a hard path:
+      # `/usr/lib/aarch64-linux-gnu/tegra-egl/libEGL_nvidia.so.0`
+      # This causes loading errors for libargus applications and the nvargus-daemon.
+      # Errors will look like this:
+      # SCF: Error NotSupported: Failed to load EGL library
+      # To fix this, create a symlink to the correct EGL library in the above directory.
+      #
+      # An alternative approach would be to wrap the library with an LD_PRELOAD to a dlopen call
+      # that replaces the hardcoded path with the correct path.
+      # However, since dynamic library symbol lookups start with the calling binary,
+      # this override would have to happen at the binary level, which means every binary
+      # would need to be wrapped. This is less desirable than simply adding the following symlink.
+      systemd.services.create-libegl-symlink =
+        let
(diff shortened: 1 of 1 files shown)


modules/default.nix
```diff
@@ -408,6 +408,33 @@ in
       # Include nv_tegra_release, just so we can tell what version our NixOS machine was built from.
       environment.etc."nv_tegra_release".source = "${pkgs.nvidia-jetpack.l4t-core}/etc/nv_tegra_release";
 
+      # For some unknown reason the libnvscf.so library has a dlopen call to a hard path:
+      # `/usr/lib/aarch64-linux-gnu/tegra-egl/libEGL_nvidia.so.0`
+      # This causes loading errors for libargus applications and the nvargus-daemon
+      # Errors will look like this:
+      # SCF: Error NotSupported: Failed to load EGL library
+      # To fix this create a symlink to the correct EGL lib at the above directory.
```
elliotberman: Also talk about how alternatives like LD_PRELOAD are difficult because it requires action from everyone who wants to (indirectly) use libnvscf.so? You spent good amount of effort to avoid the symlink, give yourself some credit

modules/default.nix
```diff
@@ -408,6 +408,33 @@ in
...
+      systemd.services.create-libegl-symlink =
+        let
+          linkEglLib = pkgs.writeShellScriptBin "link-egl-lib" ''
+            ${lib.getExe' pkgs.coreutils "mkdir"} -p /usr/lib/aarch64-linux-gnu/tegra-egl
+            ${lib.getExe' pkgs.coreutils "ln"} -s /run/opengl-driver/lib/libEGL_nvidia.so.0 /usr/lib/aarch64-linux-gnu/tegra-egl/libEGL_nvidia.so.0
+          '';
+        in
+        {
+          enable = true;
```
elliotberman, line 431: You'll have to remind me: can applications realistically use libargus without nvargus-daemon? If nvargus-daemon always required, could we move this service under modules/nvargus-daemon.nix and also guarded by services.nvargus-daemon.enable? Please also make sure this service starts before nvargus-daemon starts

sm1ll3r: Nope the libargus applications don't need the nvargus-daemon. The daemon is basically just a wrapper around the libargus api calls so the applications just call those api's directly without the need for the daemon.

modules/default.nix
sm1ll3r: Added the alternative approach. Do you think we need an enable flag around this probably don't want this on every closure. Any idea on a name for the option if so?

PR #358, opened 2025-09-11
Extend devicetrees and out-of-tree kernel modules

elliotberman:
###### Description of changes

When using hardware.deviceTree.overlays, it is useful to reference dt-bindings from JetPack. With this change, you can now:

include <dt-bindings/gpio/tegra234-gpio.h>

and other downstream JetPack headers.

Also, support including `#include <nvidia/conftest.h>` from other external kernel modules.

###### Testing

- [x] Compile a proof-of-concept overlay against orin-agx-devkit (jetpack5)
- [x] Compile a proof-of-concept overlay against orin-agx-devkit (jetpack6)

### DIFF
--- a/pkgs/kernels/r36/oot-modules.nix
+++ b/pkgs/kernels/r36/oot-modules.nix
@@ -93,6 +93,9 @@ stdenv.mkDerivation (finalAttrs: {
   postInstall = ''
     mkdir -p $dev
     cat **/Module.symvers > $dev/Module.symvers
+
+    mkdir -p $dev/include/nvidia
+    install -m 0644 out/nvidia-conftest/nvidia/conftest.h $dev/include/nvidia/
   '';
 
   outputs = [


modules/default.nix
```diff
@@ -298,6 +298,18 @@ in
...
+        "5" = let cfg = config.hardware.deviceTree; in [
+          "${lib.getDev cfg.kernelPackage}/lib/modules/${cfg.kernelPackage.modDirVersion}/source/nvidia/soc/tegra/kernel-include"
+        ];
+        # See DTC_INCLUDE inside ${gitRepos."kernel-devicetree"}/generic-dts/Makefile
+        "6" = let t23x = pkgs.nvidia-jetpack.gitRepos."hardware/nvidia/t23x/nv-public"; in [
+          "${t23x}/include/kernel"
+          "${t23x}/include/nvidia-oot"
+          "${t23x}/include/platforms"
+          "${t23x}"
+        ];
```
hacker1024: Our board vendor's BSP includes modifications to these files. It'd be great if there was a nice way to override these sources.

That being said, the existing `devicetree` package already has a similar problem, so these two occurrences can probably be changed together in a future PR.

hacker1024: Actually, upon further inspection, the `devicetree` package can be modified quite easily by replacing the `src`.

It might be more appropriate if these sources are derived from the `devicetree` package in some way.

modules/default.nix
```diff
@@ -298,6 +298,18 @@ in
       ];
 
       hardware.deviceTree.enable = true;
+      hardware.deviceTree.dtboBuildExtraIncludePaths = {
```
hacker1024, line 298: Another option here is to wrap `pkgs.deviceTree.compileDTS` in an overlay instead. This is ultimately [used by the `hardware.deviceTree` module](https://github.com/NixOS/nixpkgs/blob/e9f00bd893984bc8ce46c895c3bf7cac95331127/nixos/modules/hardware/device-tree.nix#L104-L107), but can also be used in other places - for example, we've been using it to compile DTBOs for `hardware.nvidia-jetpack.flashScriptOverrides.additionalDtbOverlays`.

Here's what we've been using on JetPack 5 (will need updating for the JP 6 paths):

```nix
{
  nixpkgs.overlays = [
    (self: super: {
      deviceTree = super.deviceTree // {
        compileDTS = args: super.deviceTree.compileDTS (args // {
          includePaths = args.includePaths or [ ] ++ lib.optionals (config.system.build.kernel != null) [
            "${lib.getDev config.system.build.kernel}/lib/modules/${config.system.build.kernel.modDirVersion}/source/include"
          ];
        });
      };
    })
  ];
}
```

elliotberman: Device trees can be compiled from sources outside the kernel tree, so I'd prefer not to make the assumption that you'll want to include kernel sources in all cases of `compileDTS`. We do set `hardware.deviceTree` to point to a kernel-built devicetree, so it makes sense for us to add the extra include paths from the kernel.

This mirrors how nixpkgs operates as well:
https://github.com/NixOS/nixpkgs/blob/90b8554d0e4cc8f7218d5e4b66e683ba2dfc1483/nixos/modules/hardware/device-tree.nix#L93-L109

modules/default.nix
elliotberman: > It might be more appropriate if these sources are derived from the devicetree package in some way.

Ack, that makes sense. I'll do that shortly

PR #361, opened 2025-09-22
optee: PKCS11 ta and xtest

TanelDettenborn:
This pull request is basically refactoring PR #149. #149 did not caught any comments and therefore this PR have a different approach: Minimal and avoids changes to existing targets.

###### Description of changes

OP-TEE has a support for PKCS#11, but currently OP-TEE's PKCS#11 support is not working within Jetpack-nixos project due it is missing required [PKCS#11 TA](https://github.com/OP-TEE/optee_os/tree/master/ta/pkcs11). This pull request adds PKCS#11 derivation and option for enabling it.

Also OP-TEE project has a sanity testsuite (xtest). Pull request adds xtest derivation and option for enabling it.

Projects, which uses Jetpack-nixos as a input, can enable PKCS11 support and xtest with option, but unfortunately pkcs11 support nor xtest has not been integrated into iso_minimal-targets. It would require a bit more extensive changes into flake.nix-file.   

###### Testing
Xtest ran on Orin NX and following tests fails:
- 1008: Loads corrupt TA. Requires investigation. It seems like the test requires write permission to TA-directory, but TAs are in nix store.
- 1033: Plugin test case. Plugin support can be disabled or passing an extra argument to tee-supplicant

### DIFF
--- a/modules/flash-script.nix
+++ b/modules/flash-script.nix
@@ -244,6 +244,22 @@ in
             };
           };
 
+          pkcs11Support = mkOption {
+            type = types.bool;
+            default = false;
+            description = ''
+              Adds OP-TEE's PKCS#11 TA.
+            '';
+          };
+
+          xtest = mkOption {
+            type = types.bool;
+            default = false;
+            description = ''
+              Adds OP-TEE's xtest and related TA/Plugins
+            '';
+          };
+
           patches = mkOption {
             type = types.listOf types.path;
             default = [ ];
--- a/mk-overlay.nix
+++ b/mk-overlay.nix
@@ -85,11 +85,11 @@ makeScope final.newScope (self: {
   );
 
   inherit (final.callPackages ./pkgs/uefi-firmware/r${l4tMajorVersion} { inherit (self) l4tMajorMinorPatchVersion; })
-    edk2-jetson uefi-firmware;
+    uefi-firmware;
 
   inherit (final.callPackages ./pkgs/optee {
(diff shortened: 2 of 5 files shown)


modules/flash-script.nix
```diff
@@ -244,6 +244,22 @@ in
             };
           };
 
+          pkcs11support = mkOption {
```
Princemachiavelli: I think proper style is `pkcs11Support`.

TanelDettenborn: Done. Also for consistency I have changed `pkcs11ta` -> `pkcs11Ta`.

Princemachiavelli:
PKCS11 works using  `pkcs11-tool` from `opensc`:
```
$ pkcs11-tool --module /nix/store/XXXX-optee_client-35.6.2/lib/libckteec.so --init-token --label "test"
$ pkcs11-tool --module /nix/store/XXXX-optee_client-35.6.2/lib/libckteec.so -L
Available slots:
Slot 0 (0x0): OP-TEE PKCS11 TA - TEE UUID XXXX
  token label        : test
  token manufacturer : Linaro
  token model        : OP-TEE TA
  token flags        : login required, rng, token initialized
  hardware version   : 0.0
  firmware version   : 0.1
  serial num         : 0000000000000000
  pin min/max        : 4/128
  uri                : pkcs11:model=OP-TEE%20TA;manufacturer=Linaro;serial=0000000000000000;token=test
```
@elliotberman This is good to go.
Thanks @TanelDettenborn !

Princemachiavelli (approved):

PR #366, opened 2025-09-30
Orin NX: Add support for xavier NX carrier board

TanelDettenborn:
Currently Jetson 35.x and 36.x BSP includes configuration/support for the original Xavier NX carrier board and Orin NX SOM:                                                                                         
 * 35.x: p3509-a02+p3767-0000.conf                                                                                                                                                                                  
 * 36.x: p3509-a02-p3767-0000.conf                                                                                                                                                                                  
                                                                                                                                                                                                                    
Jetpack-nixos has acknowdlege the fact with a comment in modules/devices.nix-file: https://github.com/anduril/jetpack-nixos/blob/master/modules/devices.nix#L118-L119                                               
                                                                                                                                                                                                                    
###### Description of changes                                                                                                                                                                                       
                                                                                                                                                                                                                    
PR introduces an option for selecting a new carrier board for the Orin NX: `xavierNxDevkit`.                                                                                                                          
                                                                                                                                                                                                                    
PR also adds `flash-orin-nx-xavierNxDevkit-jp5` and `flash-orin-nx-xavierNxDevkit` flash targets.                                                                                                                       
                                                                                                                                                                                                                    
<!--                                                                                                                                                                                                                
What has changed as a result of this PR? Why was the change made?                                                                                                                                                   
-->                                                                                                                                                                                                                 
                                                                                                                                                                                                                    
###### Testing                                                                                                                                                                                                      
Both new flash targets evaluates, flashes and boots on Orin NX SOM with original Xavier NX carrier board.


brianmcgillion (approved):

modules/default.nix
```diff
@@ -123,6 +123,9 @@ in
         type = types.enum [
           "generic"
           "devkit"
+
+          # Use this, if you would like to use Orin NX SOM with the original Xavier NX Devkit module (p3509-a02),
+          "xavierNxDevkit"
```
elliotberman, line 97: I'm hesitant about this because I'd like to make it fail in an obvious that `xavier-*-xavierNxDevkit` (or orin-agx-xavierNxDevkit) isn't a thing. I'm pretty sure it will silently fall back to the regular devkit configuration. I'm not sure of a good best practice to apply here. An easy thing to do would be an assertion like we have done for JetPack major versions.

TanelDettenborn: It is very valid concern! I have added carrierBoard-SOM assertion, but I have not generalized it, because currently only one combination is required to check.

Perhaps PR could make a bit more obvious by renaming `xavierNxDevkit` to `P3509`? P3509 is a dirty name, but it would follow Nvidia naming convention and it would not be used by _accidentally_.

PR #368, opened 2025-10-01
Revert "Apply the kernelPackagesOverlay in the NixOS module"

Princemachiavelli:
This commit breaks downstream consumers of jetpack-nixos containing overlays modifying `pkgs.nvidia-jetpack.kernelPackages.devicetree` and `pkgs.nvidia-jetpack.kernelPackages.nvidia-oot-modules`.

It's a bit unclear if this commit needs to be reworked or if downstream usages need to apply their devicetree modifications in a different manner.  If downstream changes are necessary, we should provide migration documentation. For now, I believe reverting this change is the best course.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,32 +245,6 @@ format. By default, there will be a single device setup of the kind
 
 If you are using Podman, it is recommended to add a dependency to any systemd services that run podman to specify `After=nvidia-container-toolkit-cdi-generator.service`. Due to Podman's daemonless nature, this ensures that the CDI configuration files are generated prior to container start.
 
-### Using the kernel package sets
-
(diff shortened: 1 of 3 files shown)


elliotberman:
@hacker1024 , we've had to revert this since it breaks the ability to apply overlays to NVIDIA modules/devicetree, which can be required when using non-devkit boards. I think your changes are still fundamentally good and I'd like to get them back in soon. I'm working on a [PR](https://github.com/NixOS/nixpkgs/pull/448069) to nixpkgs which will allows us to re-apply your changes.

elliotberman (approved):

PR #369, opened 2025-10-04
ota-check-firmware: Print details about why capsule update might have failed

elliotberman:
###### Description of changes

UEFI reports some details about why a capsule update might have failed. Parse these details and report it to user. Additionally, when a capsule update fails, the OS must clear the error before future capsule updates can be applied. When running ota-check-firmware, offer to the user to clear the variable.

Sample output:

```
================================= Version Info =================================
Current firmware version is : 35.6.2-d38eb20c0441
Expected firmware version is: 35.6.2-d38eb20c0441
TegraPlatformCompatSpec     : 3701-300-0000--1--jetson-agx-orin-devkit-
TegraPlatformSpec           : 3701-500-0000-H.0-1-0-jetson-agx-orin-devkit-

=============================== Boot Chain Info ================================
BootChainFwCurrent               : A
BootChainFwNext                  : Variable Doesn't Exist
BootChainFwStatus                : Variable Doesn't Exist
Capsule Update Pending           : no
ESRT capsule_flags               : 0x0
ESRT fw_class                    : bf0d4599-20d4-414e-b2c5-3595b1cda402
ESRT fw_type                     : ESRT_FW_TYPE_SYSTEMFIRMWARE
ESRT fw_version                  : 35.6.2
ESRT last_attempt_status         : SUCCESS
ESRT last_attempt_version        : 0.0.0
ESRT lowest_supported_fw_version : 35.6.2
```

###### Testing

All testing done on JP5 with orin-agx-devkit
 - [x] normal boot (output as above)
 - [x] check after successful capsule update
 - [ ] check after failed capsule update


pkgs/ota-utils/ota_helpers.sh
```diff
@@ -119,3 +119,53 @@ set_efi_var() {
...
+# Call detect_can_write_runtime_uefi_vars before running this
+rm_efi_var() {
+  local name=$1
+  local value=$2
+
+  local filepath
+
+  if [[ -n "$noRuntimeUefiWrites" ]]; then
+    if ! mountpoint -q "$espDir"; then
+      echo "$espDir is not mounted. Unable to set EFI variable."
```
danielfullmer: s/set/remove/

pkgs/ota-utils/ota-apply-capsule-update.sh
```diff
@@ -1,6 +1,14 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+# Check for the EFI Variable BootChainFwStatus. This variable is present if and
+# only if the previous capsule update failed.
```
danielfullmer: That's not correct, it's "only if", not "if".  There can be other reasons a capsule update fails that don't set the BootChainFwStatus.

pkgs/ota-utils/ota_helpers.sh
```diff
@@ -119,3 +121,72 @@ set_efi_var() {
...
+      rm "$filepath"
+    fi
+  fi
+}
+
+get_efi_str() {
+  local variable=$1
+  local p="/sys/firmware/efi/efivars/$variable"
+
+  if [[ -n "$noRuntimeUefiWrites" ]]; then
```
danielfullmer, line 160: noRuntimeUefiWrites doesn't mean we can't _read_ UEFI variables from /sys/firmware/efi/efivars.  Isn't this missing the case where we try to read a variable that we haven't already tried to write? Therefore, we need to read from /sys/firmware/efi/efivars?

danielfullmer: Actually, it's not clear to me that we want to read the "state" of a UEFI var from /EFI/NVDA/Variables like this.  If i'm reading `ota-check-firmware` output, I generally want to see what the actual current variables are--not what we're trying to set them to on the next reboot.  For instance, imagine something is wrong with the system and those variables are not actually taking effect--it'd rather have `ota-check-firmware` report what the variables actually currently are.

elliotberman: We read variable from /opt/nvidia/esp only if it exists, otherwise we read from /sys/firmware/efi. I'll add a comment to explain

danielfullmer: Ah, sorry, I missed the [[ -e ... ]] part on line 162

PR #379, opened 2025-10-17
fix(overlay): override cudaPackages scope instead of replacing it

bryanhonof:
###### Description of changes

I'd like to be able to use some of the packages that are still coming from the normal nixpkgs' cudaPackages, like cudnn-frontend, which doesn't seem to be available in the nvidia-jetpack*.cudaPackages, but do seem to compile on aarch64 and jetson nonetheless.

Just replacing the whole cudaPackages attribute set with that of the jetpack provided one feels like it might be removing packages it shouldn't. Instead, I overrideScope the original cudaPackages, and only overlay the packages that nvidia-jetpack actually provides.

I am unsure if the is the right approach, or if there's maybe a better solution. But this seems to be working?

###### Testing

Built `.legacyPackages.aarch64-linux.cudaPackages.cudnn-frontend` after overlaying it on top of the nixpkgs' nixos-25.05 branch.

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -89,7 +89,7 @@ in
   cudaPackages_11_4 =
     if redistSystem == "linux-aarch64" then
       assert final.nvidia-jetpack5.cudaPackages.cudaMajorMinorVersion == "11.4";
-      final.nvidia-jetpack5.cudaPackages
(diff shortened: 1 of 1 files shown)


bryanhonof:
Closing for now as I better understand the issue.

PR #380, opened 2025-10-17
r35: kernel: Also install dtbos in "make dtbs_install"

elliotberman:
###### Description of changes

dtbos from the kernel build are missing in the result output.

###### Testing

- [x] build r35 kernel and observe dtbos now present in the result/dtbs.


danielfullmer (approved):
It would be good to indicate why this is only needed on JP5 somewhere

PR #391, opened 2025-11-11
CUDA 13

elliotberman:
###### Description of changes

Add support for CUDA 13 for Thor AGX on JetPack 7.

Note: CUDA 13 requires a nixpkgs which has a CUDA 13 package set. Only the unstable (master) nixpkgs branch currently has the support.

###### Notes
This PR can be merged, but will fail to build a few packages against CUDA 13 without these prerequisites:
 - nixos/nixpkgs#460989 to build most of the samples
 - nixos/nixpkgs#460987 to run some of the cuda-samples
 - A patch to pkgs/top-level/cuda-packages.nix to update the CUDA 13 tensorrt version to 10.13.3 (this version is present in cuda-legacy, but not in upstream nixpkgs yet)

###### Testing

- [x] combined-test on thor-agx-devkit
- [x] combined-test on orin-agx-devkit-jp6 (to catch regressions)


elliotberman:
Depends on #389 and #390

modules/default.nix
```diff
@@ -236,8 +239,8 @@ in
               message = "JetPack NixOS 6 supports CUDA 12.4 (natively) - 12.9 (with `cuda_compat`): `pkgs.cudaPackages` has version ${cudaMajorMinorVersion}.";
             }
             {
-              assertion = !cudaAtLeast "13.0";
-              message = "JetPack NixOS does not support CUDA 13.0 or later: `pkgs.cudaPackages` has version ${cudaMajorMinorVersion}.";
+              assertion = cfg.majorVersion == "7" -> cudaAtLeast "13.0";
+              message = "JetPack NixOS 7 suppoers CUDA 13.0 (natively): `pkgs.cudaPackages` has version ${cudaMajorMinorVersion}.";
```
elliotberman:
```suggestion
              message = "JetPack NixOS 7 supports CUDA 13.0 (natively): `pkgs.cudaPackages` has version ${cudaMajorMinorVersion}.";
```

elliotberman:
> Did a read through just now and this looks good to me -- how/what are you testing?

I've been running [combined-test](https://github.com/elliotberman/jetpack-nixos/blob/cuda-13/pkgs/samples/combined-test.nix). For CUDA, I think the most relevant part is the [cuda-test](https://github.com/elliotberman/jetpack-nixos/blob/cuda-13/pkgs/samples/cuda-test.nix), running cuda-samples.

modules/default.nix
```diff
@@ -274,10 +277,14 @@ in
         "algif_skcipher"
       ];
 
-      boot.kernelModules = [ "nvgpu" ];
+      boot.kernelModules = (if (jetpackAtLeast "7") then [ "nvidia-uvm" ] else [ "nvgpu" ]);
```
danielfullmer: Remove extraneous parens around this.

modules/default.nix
```diff
@@ -288,7 +295,9 @@ in
         cudaPackages.vpi-firmware # Optional, but needed for pva_auth_allowlist firmware file used by VPI2
       ] ++ lib.optionals (l4tOlder "38") [
         l4t-xusb-firmware # usb firmware also present in linux-firmware package, but that package is huge and has much more than needed
-      ];
+      ] ++ lib.optionals (l4tAtLeast "38") (nvidiaDriverFirmwareDebs ++ [ l4t-firmware-openrm ]);
```
danielfullmer: The getDriverDebs and nvidiaDriverFirmwareDebs are only used here. Can you make a local `let` block here defining them instead of putting them in the top let block for the module?  Just keeps functionality local to the thing that uses it to reduce cognitive load.

pkgs/l4t/l4t-opencv.nix
```diff
@@ -0,0 +1,41 @@
...
+, glib
+, gst_all_1
+, gtk2
+, gtk3
+, libjpeg8
+, libunwind
+, onetbb
+, orc
+, zstd
+,
```
danielfullmer, line 14: extra unneeded `,` here.

pkgs/l4t/l4t-opencv.nix
```diff
@@ -0,0 +1,41 @@
+{ buildFromDebs
```
danielfullmer, line 1: This is future work, but we should really strive to build opencv from source if at all possible.

danielfullmer (approved):
Small nitpicks on style/organization, but LGTM

PR #398, opened 2025-11-16
Fix CI for nixpkgs master

danielfullmer:
###### Description of changes

Make CI work again....

Previously, all example nixosConfigurations were cross-compiled, but full cross-compiled systems currently have major evaluation time regressions against nixpkgs master, and aren't supported for cuda packages.

###### Testing

CI Only


elliotberman (approved):
LGTM! Send it

PR #399, opened 2025-11-16
Rearrange NixOS modules

danielfullmer:
###### Description of changes

The default.nix in the modules directory was becoming overwhelmingly complicated. This PR splits out functionality into separate files and clean things up a bit. No functional change is expected as a result of this PR.

###### Testing

CI Only

### DIFF
--- a/modules/cuda.nix
+++ b/modules/cuda.nix
@@ -71,40 +71,5 @@ in
         in
         lib.optionals (isXavier || isGeneric) [ "7.2" ] ++ lib.optionals (isOrin || isGeneric) [ "8.7" ];
     });
-
-
-
-    # For some unknown reason, the libnvscf.so library has a dlopen call to a hard path:
-    # `/usr/lib/aarch64-linux-gnu/tegra-egl/libEGL_nvidia.so.0`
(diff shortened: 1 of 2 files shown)


modules/cuda.nix
```diff
@@ -0,0 +1,110 @@
...
+        description = "Create a symlink for libEGL_nvidia.so.0 at /usr/lib/aarch64-linux-gnu/tegra-egl/";
+        unitConfig = {
+          ConditionPathExists = "!/usr/lib/aarch64-linux-gnu/tegra-egl/libEGL_nvidia.so.0";
+        };
+        serviceConfig = {
+          type = "oneshot";
+          ExecStart = lib.getExe linkEglLib;
+        };
+        wantedBy = [ "multi-user.target" ];
+      };
```
elliotberman: This should be with nvargus-daemon.nix

danielfullmer: Done.  It should really probably be replaced with a systemd-tmpfiles entry if we don't have a non-hack fix for this...

danielfullmer: In a future PR. CC @sm1ll3r

PR #405, opened 2025-11-19
Fix hardcoded dlopen's in l4t-camera

sm1ll3r:
###### Description of changes

The original fix for hardcoded dlopen's inside of Nvidia proprietary lib's was to drop a symlink using systemd.services.create-libegl-symlink. Although this fixed the issue it's not describable since it doesn't scale and ends up with a bunch of random symlinks inside /usr/lib/*

The improved fix creates a shim for dlopen and edits the symbol table to intercept any dlopen calls and replaces the bad hardcoded paths with the desired paths.

###### Testing

Tested using nvraw on an orin-agx. Saw the hardcoded path get redirected to the correct path.

```
nvargus_nvraw version 1.16.0
Redirecting /usr/lib/aarch64-linux-gnu/tegra-egl/libEGL_nvidia.so.0 to /run/opengl-driver/lib/libEGL_nvidia.so.0
+------ Exposure info results -----------+
| Exposure | Exposure Time | Sensor Gain |
|        0 |      0.038114 |     1.55655 | 
+----------------------------------------+
NvMMLiteBlockCreate : Block : BlockType = 1 
writeToFile: file sample.jpg is successfully created
writeToFile: file sample.nvraw is successfully created
```

Also added a small sanity test to ensure the override works as expected.

### DIFF
--- a/mk-overlay.nix
+++ b/mk-overlay.nix
@@ -99,7 +99,7 @@ makeScope final.newScope (self: {
 
   buildFromDebs = final.callPackage ./pkgs/buildFromDebs.nix { inherit (self) debs cudaMajorMinorVersion defaultSomDebRepo; };
 
-  dlopenOverride = final.callPackage ./pkgs/dlopenOverride.nix { };
+  dlopenOverride = final.callPackage ./pkgs/dlopen-override { };
 
   inherit (final.callPackages ./pkgs/uefi-firmware/r${l4tMajorVersion} { inherit (self) l4tMajorMinorPatchVersion; })
     uefi-firmware;
--- a/pkgs/dlopen-override/default.nix
+++ b/pkgs/dlopen-override/default.nix
@@ -3,8 +3,18 @@
 , buildPackages
 }:
 
-# rewrites is an attribute set of the form
+# For some unknown reason multiple Nvidia libraries hard code dlopen paths.
+# For example the libnvscf.so library has a dlopen call to a hard path:
+# `/usr/lib/aarch64-linux-gnu/tegra-egl/libEGL_nvidia.so.0`
+# This causes loading errors for libargus applications and the nvargus-daemon.
+# Errors will look like this:
+# SCF: Error NotSupported: Failed to load EGL library
+#
+# To fix this, we are creating a dlopen shim that replaces the hard coded path with a known good path
+#
+# To use this pass rewrites which is an attribute set of the form
 # { oldPath = newPath; }
+# And a file which is just a string to the exact path of the lib
 rewrites: file:
 let
   oldPaths = builtins.attrNames rewrites;
@@ -16,7 +26,7 @@ let
 
     strictDeps = true;
 
-    src = ./dlopen-override;
+    src = ./.;
 
     patchPhase = ''
       substituteInPlace dlopenoverride.c \
@@ -49,10 +59,13 @@ let
   };
 in
 ''
-  remapFile=$(mktemp)
-  echo dlopen ${dlopenUnique} > $remapFile
-  ${lib.getExe buildPackages.patchelfUnstable} ${file} \
-    --rename-dynamic-symbols "$remapFile" \
-    --add-needed ${dlopen}/lib/dlopen-override.so \
-    --add-rpath ${dlopen}/lib
+  (
+    remapFile=$(mktemp)
+    echo dlopen ${dlopenUnique} > $remapFile
+    ${lib.getExe buildPackages.patchelfUnstable} ${file} \
+      --rename-dynamic-symbols "$remapFile" \
+      --add-needed ${dlopen}/lib/dlopen-override.so \
+      --add-rpath ${dlopen}/lib
+    rm $remapFile
+  )
 ''
--- a/pkgs/dlopen-override/dlopenoverride.c
+++ b/pkgs/dlopen-override/dlopenoverride.c
@@ -1,6 +1,7 @@
 #include <dlfcn.h>
 #include <stdio.h>
 #include <string.h>
+#include <stdlib.h>
 
 // List of paths to replace
 const char *old_paths[] = { "@oldpaths@" };
@@ -8,10 +9,14 @@ const char *new_paths[] = { "@newpaths@" };
 const int num_paths = sizeof(old_paths) / sizeof(old_paths[0]);
 
 void *__dlopen(const char *filename, int flag) {
+    char *debug = getenv("DEBUG");
+
     if (filename) {
         for (int i = 0; i < num_paths; i++) {
             if (strcmp(filename, old_paths[i]) == 0) {
-                fprintf(stderr, "Redirecting %s to %s\n", filename, new_paths[i]);
+                if (debug != NULL) {
+                    fprintf(stderr, "Redirecting %s to %s\n", filename, new_paths[i]);
+                }
                 filename = new_paths[i];
                 break;
             }
--- a/pkgs/l4t/l4t-camera.nix
+++ b/pkgs/l4t/l4t-camera.nix
@@ -14,6 +14,8 @@ buildFromDebs {
     ln -srfv lib/libv4l2_nvargus.so lib/libv4l/plugins/nv/libv4l2_nvargus.so
   '';
 
+  # Unfortunately we have to do this dlopenOverride in postFixupHooks because autoPathElf and addDriverRunpaths override
+  # the needed and rpath sections of the binary leading to our changes getting removed.
   preFixup = ''
     postFixupHooks+=('
       ${ dlopenOverride { "/usr/lib/aarch64-linux-gnu/tegra-egl/libEGL_nvidia.so.0" = "/run/opengl-driver/lib/libEGL_nvidia.so.0"; } "$out/lib/libnvscf.so" }
--- a/flake.nix
+++ b/flake.nix
@@ -183,6 +183,9 @@
           inherit lib nixpkgs pkgs system;
           overlay = self.overlays.default;
         };
+        dlopenOverride = pkgs.runCommand "dlopen-override-test" { } ''
+          ${lib.getExe self.legacyPackages.${system}.tests.dlopen-override} && touch $out
+        '';
       });
 
       formatter = forAllSystems ({ pkgs, ... }: import ./treefmt.nix pkgs);


pkgs/dlopenOverride.nix
```diff
@@ -0,0 +1,50 @@
+{ stdenv
+, gcc
+, writeText
+, lib
+, buildPackages
+}:
+
+
+oldPaths: newPaths: file:
```
sm1ll3r: Should probably add an assert that oldPaths and newPaths are the same length?

pkgs/dlopen-override/dlopenoverride.c
```diff
@@ -0,0 +1,22 @@
...
+// List of paths to replace
+const char *old_paths[] = { "@oldpaths@" };
+const char *new_paths[] = { "@newpaths@" };
+const int num_paths = sizeof(old_paths) / sizeof(old_paths[0]);
+
+void *__dlopen(const char *filename, int flag) {
+    if (filename) {
+        for (int i = 0; i < num_paths; i++) {
+            if (strcmp(filename, old_paths[i]) == 0) {
+                printf("Redirecting %s to %s\n", filename, new_paths[i]);
```
elliotberman: print to stderr since stdout might be precious

pkgs/dlopenOverride.nix
elliotberman: What if we make the input a attribute set. `{ old = new; }`, then:
```
let
    oldPaths = builtins.attrNames rewrites;
    newPaths = builtins.attrValues rewrites;
in
```

pkgs/dlopenOverride.nix
```diff
@@ -0,0 +1,50 @@
...
+let
+  dlopenHash = builtins.hashString "sha256" ((lib.concatStrings oldPaths) + (lib.concatStrings newPaths) + file);
```
ConnorBaker: Adding to Elliot's comment: if you use `rewrites` as the mapping between old and new paths, since Nix canonicalizes attribute sets before converting to JSON, you can do 


```suggestion
  dlopenHash = builtins.hashString "sha256" (builtins.toJSON rewrites);
```

Notably, `file` is left out of it -- `file` isn't involved in the creation of `dlopen`, so it shouldn't be part of the `dlopenHash`.

pkgs/dlopen-override/default.nix
```diff
@@ -0,0 +1,50 @@
...
+, writeText
+, lib
+, buildPackages
+}:
+
+
+oldPaths: newPaths: file:
+let
+  dlopenHash = builtins.hashString "sha256" ((lib.concatStrings oldPaths) + (lib.concatStrings newPaths) + file);
+  dlopen = stdenv.mkDerivation {
```
ConnorBaker, line 23: Enable `strictDeps = true;`.

pkgs/dlopenOverride.nix
sm1ll3r: Done!

pkgs/dlopenOverride.nix
```diff
@@ -0,0 +1,50 @@
...
+  dlopen = stdenv.mkDerivation {
+    pname = "dlopen-override";
+    version = "1.0";
+
+    src = ./dlopen-override;
+
+    buildInputs = [ gcc ];
```
ConnorBaker: A compiler is provided by `stdenv`, no need to specify it as an input.

pkgs/dlopen-override/dlopenoverride.c
sm1ll3r: Done

pkgs/dlopenOverride.nix
```diff
@@ -0,0 +1,50 @@
...
+          --replace-fail \
+            '@newpaths@' \
+            '${ builtins.concatStringsSep "\",\"" newPaths }' \
+          --replace-fail \
+            '__dlopen' \
+            '__dlopen${dlopenHash}'
+    '';
+
+    buildPhase = ''
+      gcc -shared -fPIC -o dlopen-override.so dlopenoverride.c -ldl
```
ConnorBaker:
```suggestion
      cc -shared -fPIC -o dlopen-override.so dlopenoverride.c -ldl
```

pkgs/dlopenOverride.nix
```diff
@@ -0,0 +1,50 @@
...
+    '';
+
+    installPhase = ''
+      mkdir -p $out/lib
+      cp dlopen-override.so $out/lib/
```
ConnorBaker:
```suggestion
      install -Dm755 dlopen-override.so "$out"/lib/
```

sm1ll3r: Done

pkgs/dlopenOverride.nix
sm1ll3r: Done

pkgs/dlopenOverride.nix
sm1ll3r: Done

pkgs/dlopen-override/default.nix
sm1ll3r: Done

pkgs/dlopenOverride.nix
sm1ll3r: Done

mk-overlay.nix
```diff
@@ -256,7 +252,7 @@ makeScope final.newScope (self: {
     inherit (finalSamples) callPackage;
   });
 
-  tests = final.callPackages ./pkgs/tests { inherit l4tMajorMinorPatchVersion l4tAtLeast; };
+  tests = self.callPackages ./pkgs/tests { inherit l4tMajorMinorPatchVersion l4tAtLeast; };
```
elliotberman:
```suggestion
  tests = self.callPackages ./pkgs/tests { };
```

pkgs/dlopenOverride.nix
```diff
@@ -0,0 +1,58 @@
...
+, buildPackages
+}:
+
+# rewrites is an attribute set of the form
+# { oldPaths = newPaths; }
+rewrites: file:
+let
+  oldPaths = builtins.attrNames rewrites;
+  newPaths = builtins.attrValues rewrites;
+  dlopenHash = builtins.hashString "sha256" (builtins.toJSON rewrites);
```
elliotberman:
```suggestion
  dlopenUnique = "__dlopen.${builtins.hashString "sha256" (builtins.toJSON rewrites)}";
```

pkgs/dlopenOverride.nix
```diff
@@ -0,0 +1,58 @@
...
+      substituteInPlace dlopenoverride.c \
+          --replace-fail \
+            '@oldpaths@' \
+            '${ builtins.concatStringsSep "\",\"" oldPaths }' \
+          --replace-fail \
+            '@newpaths@' \
+            '${ builtins.concatStringsSep "\",\"" newPaths }' \
+          --replace-fail \
+            '__dlopen' \
+            '__dlopen${dlopenHash}'
```
elliotberman:
```suggestion
            '${dlopenUnique}'
```

pkgs/dlopenOverride.nix
```diff
@@ -0,0 +1,58 @@
...
+
+      install -Dm755 dlopen-override.so "$out"/lib/dlopen-override.so
+
+      runHook postInstall
+    '';
+  };
+in
+''
+  remapFile=$(mktemp)
+  echo dlopen __dlopen${dlopenHash} > $remapFile
```
elliotberman:
```suggestion
  echo dlopen ${dlopenUnique} > $remapFile
```

pkgs/dlopenOverride.nix
```diff
@@ -0,0 +1,58 @@
+{ stdenv
+, lib
+, buildPackages
+}:
+
+# rewrites is an attribute set of the form
+# { oldPaths = newPaths; }
```
elliotberman:
```suggestion
# { oldPath = newPath; }
```

elliotberman (approved):
some nits, but looks good

pkgs/l4t/l4t-camera.nix
```diff
@@ -12,4 +13,11 @@ buildFromDebs {
   postPatch = ''
     ln -srfv lib/libv4l2_nvargus.so lib/libv4l/plugins/nv/libv4l2_nvargus.so
   '';
+
+  preFixup = ''
```
danielfullmer, line 19: Can you elaborate why you're doing it like this?

pkgs/dlopen-override/default.nix
```diff
@@ -0,0 +1,58 @@
...
+      runHook preInstall
+
+      install -Dm755 dlopen-override.so "$out"/lib/dlopen-override.so
+
+      runHook postInstall
+    '';
+  };
+in
+''
+  remapFile=$(mktemp)
```
danielfullmer: Need to remove remapFile after running patchelf

pkgs/dlopenOverride.nix
```diff
@@ -0,0 +1,58 @@
...
+    installPhase = ''
+      runHook preInstall
+
+      install -Dm755 dlopen-override.so "$out"/lib/dlopen-override.so
+
+      runHook postInstall
+    '';
+  };
+in
+''
```
danielfullmer, line 61: Since this output is just a snippet of bash code, could you put these commands in a subshell by wrapping the whole thing in `( )` to help isolate execution behavior of this snippet from the environment of whatever potential script it's being included into.

pkgs/dlopen-override/default.nix
```diff
@@ -0,0 +1,58 @@
+{ stdenv
```
danielfullmer, line 1: move this file into `pkgs/dlopen-override/default.nix` ?

pkgs/dlopen-override/default.nix
```diff
@@ -0,0 +1,58 @@
...
+, lib
+, buildPackages
+}:
+
+# rewrites is an attribute set of the form
+# { oldPath = newPath; }
```
danielfullmer, line 16: Could you write some more comprehensive documentation about what this utility is, why it's needed, and how it achieves its intended result?

pkgs/dlopen-override/default.nix
```diff
@@ -0,0 +1,58 @@
...
+rewrites: file:
+let
+  oldPaths = builtins.attrNames rewrites;
+  newPaths = builtins.attrValues rewrites;
+  dlopenUnique = "__dlopen${builtins.hashString "sha256" (builtins.toJSON rewrites)}";
```
danielfullmer, line 22: I'm not 100% sure of the answer, just asking the question, but is the mapping of rewrites enough to uniquely identify this library?  Think about in context of multiple dlopen overrides across different libraries in a tree of shared libraries.

pkgs/dlopen-override/dlopenoverride.c
```diff
@@ -0,0 +1,22 @@
...
+// List of paths to replace
+const char *old_paths[] = { "@oldpaths@" };
+const char *new_paths[] = { "@newpaths@" };
+const int num_paths = sizeof(old_paths) / sizeof(old_paths[0]);
+
+void *__dlopen(const char *filename, int flag) {
+    if (filename) {
+        for (int i = 0; i < num_paths; i++) {
+            if (strcmp(filename, old_paths[i]) == 0) {
+                fprintf(stderr, "Redirecting %s to %s\n", filename, new_paths[i]);
```
danielfullmer: We shouldn't unconditionally print something here. You can do something where it only prints if a certain environment variable is set though.

danielfullmer (approved):
I don't think this is going to be sufficient for the case when we have multiple uses of overrides in the same tree of shared libraries.  If you did this twice, you would have two libraries, but they would both have the same DT_SONAME, and I believe only one of them would get loaded.  It'd be valuable to make a unique SONAME for each library and have a test asserting that you can include two dlopen override shims in the library loading tree of one binary.

Still approved this, but you can improve this in a follow up PR.

pkgs/dlopen-override/default.nix
sm1ll3r: Besides the SONAME issue you mention which I'll make a follow up for. I think the hash should be unique enough since the shim lib should be exactly the same if two libs need to replace the same dlopen paths.

pkgs/dlopen-override/default.nix
```diff
@@ -0,0 +1,71 @@
...
+    installPhase = ''
+      runHook preInstall
+
+      install -Dm755 dlopen-override.so "$out"/lib/dlopen-override.so
+
+      runHook postInstall
+    '';
+  };
+in
+(
```
danielfullmer: By subshell with `( )`  I meant
```
'' (
<shell code here>
) ''
```

Currently the parens you added are Nix-language parens, which in this case aren't doing anything.  If you have the parens in the shell code then they create a subshell which has more independence from the surrounding script

PR #407, opened 2025-11-19
overlay: JetPack CUDA extensions should only apply when using JetPack

elliotberman:
###### Description of changes

The CUDA extension was erroneously applied unconditionally. It should only apply when using the JetPack CUDA package sets.

Fixes: d4efffc357e0 ("cuda: Introduce L4T cuda-extensions")

###### Testing

- [x] `vpi` attribute should not exist unless on a JetPack CUDA system.


danielfullmer (approved):
Agree with this if this is what @ConnorBaker intended.

PR #408, opened 2025-11-20
Script updates for flash-from-device.sh

rlfishback:
###### Description of changes

- Added quotes around variables to prevent `sh: : unknown operand` warnings
- Added calls to function report_step to make sure variable current_step is updated every loop iteration

###### Testing

Tested on an Orin AGX industrial


davidgantman (approved):
LGTM

PR #414, opened 2025-11-29
feat: upgrade to 25.11 and re-use upstream's package expressions

ConnorBaker:
###### Description of changes

> [!NOTE]
>
> This PR updates the Nixpkgs flake input from `nixos-25.05` to `nixos-25.11`.

Since the introduction of [nixos-cuda/cuda-legacy](https://github.com/nixos-cuda/cuda-legacy) and the various changes I've made upstream to support it, we can re-use many, many more package expressions from upstream, meaning we no longer need to vendor them.

This PR essentially removes the vendored package expressions introduced as part of JetPack 6/CUDA 12 support because upstream was pending a refactor (the package expressions were largely taken from my WIP refactor branch for upstream).

As an additional change, all CUDA package sets are now created as *extensions* of the CUDA package sets available upstream (see https://github.com/NixOS/nixpkgs/blob/c5220bcad8d7e599432db935f1159a81d5019dd4/doc/languages-frameworks/cuda.section.md?plain=1#L66), meaning that we benefit from using the same setup hooks and gain the same niceties they do (e.g., `backendStdenv`).

Additionally, the NixOS module setting up graphics for Jetson devices includes many more libraries, matching the larger set of functionality included in NVIDIA's desktop drivers.

Lastly, specific changes for CUDA 11.4/JetPack 5:

- we are able to use upstream's package expressions modulo a little bit of normalization of unpacked debian installers
- libnvinfer-samples has been removed in favor of cudaPackages.tensorrt-samples
- TensorRT 8.5.2 is provided for x86_64-linux to match the version provided for Jetson

###### Testing

- [x] Verify `samples.combinded-test` works across JetPack 5/6/7
- [x] Internal HITL testing
  - [x] Xavier JP5
  - [x] Orin JP5
  - [x] Orin JP6
  - [x] Thor JP7


overlay.nix
```diff
@@ -124,48 +127,57 @@ in
...
-      assert final.nvidia-jetpack6.cudaPackages.cudaMajorMinorVersion == "12.6";
-      final.nvidia-jetpack6.cudaPackages
-    else
-      prev.cudaPackages_12_6;
-  cudaPackages_12 =
-    if useJetPackCudaPackageSet then
-      final.cudaPackages_12_6
-    else
-      prev.cudaPackages_12;
+  # Override upstream's manifest selection so the version of TensorRT used is consistent
```
elliotberman:
```suggestion
  # Override upstream's manifest selection so the version of TensorRT used is consistent with L4T
```

flake.nix
```diff
@@ -1,6 +1,6 @@
 {
   inputs = {
-    nixpkgs.url = "github:nixos/nixpkgs/nixos-25.05";
+    nixpkgs.url = "github:nixos/nixpkgs/master";
```
danielfullmer: This should target `nixos-25.11`, not master.

pkgs/l4t/l4t-3d-core.nix
```diff
@@ -50,6 +50,9 @@ buildFromDebs {
           --replace-needed libnvos.so libnvos_3d.so
       fi
     done
+
+    rm lib/ld.so.conf
```
danielfullmer: This fails to build on `thor-agx-devkit`, since `lib/ld.so.conf` doesn't exist in that deb. Just add `-f` to this and the line below.

flake.nix
ConnorBaker: Fixed in latest.

pkgs/l4t/l4t-3d-core.nix
ConnorBaker: Good catch, fixed in latest (force-pushed).

elliotberman (approved):

PR #415, opened 2025-11-30
mk-overlay.nix: fix usage of CUDA implementation details post 25.11

ConnorBaker:
###### Description of changes

Fix usage of implementation details since https://github.com/NixOS/nixpkgs/pull/462761.

###### Testing

Verified the `backendStdenv.nix` error is fixed by ensuring

```console
nix eval .#legacyPackages.aarch64-linux.nvidia-jetpack5.cudaPackages.saxpy.drvPath --override-input nixpkgs github:nixos/nixpkgs/nixos-25.05
```

and

```console
nix eval .#legacyPackages.aarch64-linux.nvidia-jetpack5.cudaPackages.saxpy.drvPath --override-input nixpkgs github:nixos/nixpkgs/nixos-25.11
```

evaluate.

Verified the missing `manifests` attribute error is fixed by ensuring

```console
nix eval .#legacyPackages.aarch64-linux.nvidia-jetpack6.cudaPackages.pkgs.cudaPackages_11_6.pkgs.nvidia-jetpack.cudaPackages.saxpy.drvPath --override-input nixpkgs github:nixos/nixpkgs/nixos-25.05
```

and

```console
nix eval .#legacyPackages.aarch64-linux.nvidia-jetpack6.cudaPackages.pkgs.cudaPackages_11_6.pkgs.nvidia-jetpack.cudaPackages.saxpy.drvPath --override-input nixpkgs github:nixos/nixpkgs/nixos-25.11
```

evaluate.

### DIFF
--- a/mk-overlay.nix
+++ b/mk-overlay.nix
@@ -178,6 +178,11 @@ makeScope final.newScope (self: {
         inherit (self) debs buildFromDebs; # NOTE: The presence of debs is used as a condition in construciton of pkgs'.
         debsForSourcePackage = srcPackageName: filter (pkg: (pkg.source or "") == srcPackageName) (attrValues finalCudaPackages.debs.common);
 
+        # The manifests attribute is required to avoid breakages caused by going from a JetPack-provided CUDA package
+        # set to one of upstream Nixpkgs' CUDA package sets, since upstream compares against the manifests attribute.
+        # See: https://github.com/NixOS/nixpkgs/pull/462761.
+        manifests.cuda.release_label = cudaMajorMinorPatchVersion;
+
         pkgs = pkgs';
 
         # Use backendStdenv from upstream


flake.nix
```diff
@@ -197,7 +197,7 @@
               inherit system;
               config = {
                 allowUnfree = true;
-                cudaCapabilities = [ "7.2" "8.7" ];
+                cudaCapabilities = [ "8.7" ];
```
danielfullmer: Is there any reason we need to have `legacyPackages` here bring in a `cudaSupport=true` version of nixpkgs?

danielfullmer: Like, it seems awkward to only support the Orin compute capability here and not Xavier.   If we're going to enable cudaSupport in the legacyPackages output i would rather we get them both working--or neither.
Users could still access these packages through `.#nixosConfigurations.<name>.pkgs`

ConnorBaker: That's my bad, I did that to verify everything evaluated (didn't mean to commit that change). Fixing now.

ConnorBaker: Okay, force pushed and should be good.

PR #426, opened 2025-12-17
R36 and "bring our own kernel"-feature related conftests fixes.

TanelDettenborn:
Jetson R36 introduces "[bring your own kernel](https://docs.nvidia.com/jetson/archives/r36.4.4/DeveloperGuide/SD/Kernel/BringYourOwnKernel.html)"-feature. Nvidia-oot requires a couple of fixes to work against upstream kernel. Note: Tested with kernel 6.6.118. Later kernel might require more adjusting of nvidia-oot.

###### Description of changes
* Nix adds "format-security" GCC flag and therefore one of the NVIDIA's "conftest"-test error erroneously.
* Display driver is missing "of_property_for_each_u32_removed_internal_args"-conftest

###### Testing
`nix build .#flash-orin-nx-debug && nix build .#iso_minimal` evaluates and compile (and boots on NX device). Targets has been tested with default BSP 5.15.x kernel and upstream 6.6.118 kernel.

### DIFF
--- a/pkgs/kernels/r36/0001-Add-of_property_for_each_u32_removed_internal_args-c.patch
+++ b/pkgs/kernels/r36/0001-Add-of_property_for_each_u32_removed_internal_args-c.patch
@@ -1,0 +1,93 @@
+From ccea708ddb09b64c83a7b08452e0c81365a49db6 Mon Sep 17 00:00:00 2001
+From: Tanel Dettenborn <tanel.dettenborn@tii.ae>
+Date: Thu, 4 Dec 2025 16:06:45 +0200
+Subject: [PATCH] Add "of_property_for_each_u32_removed_internal_args"-conftest
+
+Compiling fails against 6.6.76 and onward kernels, because internal
+arguments are removed from "of_property_for_each_u32"-macro.
+
+Note: Conftest is ported from "nvidia-oot"-repository.
+
+Signed-off-by: Tanel Dettenborn <tanel.dettenborn@tii.ae>
+---
+ kernel-open/conftest.sh                       | 23 +++++++++++++++++++
+ kernel-open/nvidia/nv-dsi-parse-panel-props.c |  6 +++++
+ kernel-open/nvidia/nvidia.Kbuild              |  1 +
+ 3 files changed, 30 insertions(+)
+
+diff --git a/kernel-open/conftest.sh b/kernel-open/conftest.sh
+index 391374f..7bdd5cb 100644
+--- a/kernel-open/conftest.sh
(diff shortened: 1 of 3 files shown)


elliotberman:
We don't have the bandwidth to "bring your own kernel". I'm happy to help take work to support BYOK, but we might inadvertently break (or be forced to break) BYOK. If we know that the patches don't apply and can't trivially fix them up, we can tag you and give some time to clean up the patches.

---

I assume there's more to add support for the 6.6 kernel than just these 2 out-of-tree module patches?

elliotberman:
If you'd like to add support for 6.6 kernel in jetpack-nixos, I'd like to bring in all the changes rather than just a subset of them.

danielfullmer:
Are these fixes still needed with the latest version of R36?   With the work-in-progress in https://github.com/anduril/jetpack-nixos/pull/487 we'd be open to fixes to the conftest code against more recent kernels.

> Lists the patches, and which kernel they landed in. depending on your baseline kernel, so if you have a 6.6 you need to apply 6.7 and next.

I think we'll probably just pick a given kernel version as our lowest supported mainline kernel version (maybe just the LTS kernel used by a given NixOS release?), and only add patches for that--at least in jetpack-nixos, users are of course free to choose their own version and patch it themselves.

TanelDettenborn:
My apologies for letting this PR go stale. I hit some implementation dilemmas and got stuck (not satfied) and ended up getting sidetracked.

>Are these fixes still needed with the latest version of R36?

Jetson BSP 36.5 requires the `0001-Fix-register_shrinker_has_fmt_arg-conftest.patch` when using kernels from 6.0.x to 6.6.x. However, the `0001-Add-of_property_for_each_u32_removed_internal_args-c.patch` is no longer needed since 36.5 now includes a similar conftest.

>I think we'll probably just pick a given kernel version as our lowest supported mainline kernel version (maybe just the LTS kernel used by a given NixOS release?),

Nixos might be using a bit too new kernel for r36? It is a question, because I have not tried to compile the latest R36 against anything newer than 6.6.129 kernel. I tried compiling older R36 releases against newer kernels, but the required changes to NVIDIA OOT just were not feasible from a maintenance standpoint.

Sticking to a single upstream LTS kernel makes sense, but the latest version might not be the most sensible choice. I would prefer to select the path of least resistance for maintenance and only bump the kernel version when Nvidia's OOT code catches up.

If you would like, I can refactor this PR to add upstream kernel support for R36. I would rebase it on top of #487 and introduce an option to toggle between BSP/upstream kernel. For the upstream version, I would target 6.6.130.

PR #429, opened 2025-12-23
l4t-3d-core: Fix the appendRunpaths attribute

elliotberman:
###### Description of changes

appendRunpaths expects the full path you want to append, and doesn't do any further manipulation. We intended to add the library path of libglvnd and others, but ended up only adding the base $out output.

Fixes: 4dfea63a39e3 ("l4t-3d-core: Use appendRunpaths")

###### Testing

- [x] run vulkaninfo

### DIFF
--- a/pkgs/l4t/l4t-3d-core.nix
+++ b/pkgs/l4t/l4t-3d-core.nix
@@ -52,5 +52,5 @@ buildFromDebs {
     done
   '';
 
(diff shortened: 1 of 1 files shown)


ConnorBaker (approved):

ConnorBaker:
Most excellent!

There shouldn't be any whitespace in any of the paths, so no need for `__structuredAttrs` to be set, right?

PR #433, opened 2026-01-16
initrdflash: Ensure secondary GPT is placed at end of disk

elliotberman:
###### Description of changes

Presently, the secondary_gpt is placed sequentially after the last partition on a disk. The secondary (backup) GPT must be at the last LBA of the disk [1]. If the partition layout doesn't fill the entire disk, then the current behavior of "placed sequentially after the last partition on a disk" is incorrect.  On all devices except AGX Xavier, this is fine because the disk is filled with other partitions such that the secondary_gpt is naturally in the right spot. Although L4T documents that setting the size property of secondary_gpt to 0xffffffffffffffff to ensure the table is placed at the end of the disk [2], TegraFlash only does this when it itself is flashing (i.e. legacy non-initrd flash method) using the storageinfo mechanism[3]. Furthermore, num_sectors in flash.xml on AGX Xavier is not correct, so build-time manipulation of flash.idx won't get us correct behavior. We are left to update the location of the secondary GPT during flashing. Tools like sgdisk make me nervous because they may modify the front of the disk or adjust partition table entries. Instead, I've opted to manually moving the secondary table and being careful to only change what we know needs to be changed. This ensures that the GPT is preserved as NVIDIA's flashing scripts expects.

To ensure that secondary_gpt is placed at the end of the disk, we re-compute the start_location on-device in the rcmboot'd kernel.

[1]: https://uefi.org/specs/UEFI/2.10/05_GUID_Partition_Table_Format.html#guid-partition-table-gpt-disk-layout-1
[2]: https://docs.nvidia.com/jetson/archives/r35.6.1/DeveloperGuide/AR/BootArchitecture/PartitionConfiguration.html?highlight=sdmmc_boot#partition-child-elements
[3]: tegraparser_v2 --storageinfo storage_info.bin --generategpt --pt flash.xml.bin

###### Testing

- [x] Run initrd flash and UEFI capsule update tests on AGX Xavier
- [x] Run initrd flash and UEFI capsule update on AGX Orin (JetPack 5 & 6)
- [x] Run initrd flash and UEFI capsule update on AGX Thor

### DIFF
--- a/pkgs/flash-from-device/flash-from-device.sh
+++ b/pkgs/flash-from-device/flash-from-device.sh
@@ -229,6 +229,12 @@ write_partitions() {
       file_size=$(stat -c "%s" "$partfile")
       if [[ -n "$disk_size" ]]; then
         start_location=$((disk_size - file_size))
+
+        # From edk2-nvidia/Silicon/NVIDIA/Include/Library/GptLib.h
+        # #define NVIDIA_GPT_BLOCK_SIZE   512
+        patchgpt "$partfile" "$start_location" 512 >"patched"
+        partfile="patched"
+        echo "Moving secondary_gpt to end: $disk_size - $file_size = $start_location"
       else
         echo "WARNING: could not ensure secondary GPT is at end of disk"
       fi
--- a/pkgs/flash-from-device/default.nix
+++ b/pkgs/flash-from-device/default.nix
@@ -20,6 +20,7 @@ let
     cp ${mtdutils}/bin/mtd_debug $out/bin
     cp ${mtdutils}/bin/flash_erase $out/bin
     cp ${tegra-eeprom-tool-static}/bin/tegra-boardspec $out/bin
+    cp ${lib.getExe pkgsStatic.nvidia-jetpack.patchgpt} $out/bin
   '';
   name = "flash-from-device";
 in


pkgs/flash-from-device/flash-from-device.sh
```diff
@@ -199,41 +216,51 @@ write_partitions() {
...
+
+    # secondary_gpt must be placed at end of disk and flash.idx wasn't generated
+    # with enough info to place it at the end
+    if [[ "$partname" == "secondary_gpt" ]] ; then
+      disk_size=$(disk_size "$devnum" "$instnum")
+      file_size=$(stat -c "%s" "$partfile")
+      if [[ -n "$disk_size" ]]; then
+        start_location=$((disk_size - file_size))
+
+        patchgpt "$partfile" "$start_location" 512 >"patched"
```
elliotberman, line 235: TODO compute block size (512)

pkgs/flash-from-device/flash-from-device.sh
```diff
@@ -199,41 +216,53 @@ write_partitions() {
...
       fi
-    else
-      report_step "Skipping flash.idx entry:$partname (devnum=$devnum, instnum=$instnum) (offset=$start_location)"
     fi
   done <flash.idx
 }
 
 find_matching_spec
 
+echo "boardspec: $matching_boardspec"
```
danielfullmer, line 264: leftover debug message?

danielfullmer (approved):

pkgs/flash-from-device/flash-from-device.sh
elliotberman: Offline discussion and concluded it's a helpful debug message to keep

PR #436, opened 2026-01-18
xavier-nx-devkit: Ensure TBCDTB_FILE is set to match DTB_FILE

elliotberman:
###### Description of changes

For the Xavier NX, the `initrd` flash script was failing to correctly write platform firmware because it assumed the target SD card would be at `/dev/mmcblk0`. However, the device tree was configured to probe an unused eMMC controller (`sdhci@3460000`) first, which resulted in the actual SD card being assigned to `/dev/mmcblk1`. This discrepancy caused the flash script to write to a non-existent device. The problem went unnoticed for some time because SD cards retained data from previous, functional flashing methods.

To resolve this, the `TBCDTB_FILE` is now set to match the `DTB_FILE`. As a result, the SD card controller (`sdhci@3440000`) is correctly identified, and the SD card is properly assigned to `/dev/mmcblk0`, which the `initrd` flash script expects.

I've updated xavier-nx-emmc as well, which removes the SD card controller as well.

###### Testing

- [x] Flash xavier-nx-devkit

### DIFF
--- a/modules/devices.nix
+++ b/modules/devices.nix
@@ -158,11 +158,13 @@ in
         (mkIf (cfg.som == "xavier-nx") {
           targetBoard = mkDefault "jetson-xavier-nx-devkit";
           partitionTemplate = mkDefault (filterPartitions defaultPartitionsToRemove "${pkgs.nvidia-jetpack.bspSrc}/bootloader/${partitionTemplateDirectory}/cfg/flash_l4t_t194_spi_sd_p3668.xml");
+          patches = [ ../pkgs/r35-bsp.patch ]; # See: https://github.com/anduril/jetpack-nixos/pull/436
         })
 
         (mkIf (cfg.som == "xavier-nx-emmc") {
           targetBoard = mkDefault "jetson-xavier-nx-devkit-emmc";
           partitionTemplate = mkDefault (filterPartitions defaultPartitionsToRemove "${pkgs.nvidia-jetpack.bspSrc}/bootloader/${partitionTemplateDirectory}/cfg/flash_l4t_t194_spi_emmc_p3668.xml");
+          patches = [ ../pkgs/r35-bsp.patch ]; # See: https://github.com/anduril/jetpack-nixos/pull/436
         })
       ];
   }
--- a/pkgs/r35-bsp.patch
+++ b/pkgs/r35-bsp.patch
@@ -1,0 +1,35 @@
+diff --git a/p3509-0000+p3668-0000-qspi-sd.conf b/p3509-0000+p3668-0000-qspi-sd.conf
+index e327f03..2c5a96a 100644
+--- a/p3509-0000+p3668-0000-qspi-sd.conf
++++ b/p3509-0000+p3668-0000-qspi-sd.conf
+@@ -38,6 +38,7 @@ source "${LDK_DIR}/p3668.conf.common";
+ BLBlockSize=1048576;
+ EMMC_CFG=flash_l4t_t194_spi_sd_p3668.xml;
+ DTB_FILE=tegra194-p3668-0000-p3509-0000.dtb;
++TBCDTB_FILE=tegra194-p3668-0000-p3509-0000.dtb
+ RECROOTFSSIZE=100MiB
+ MISC_COLD_BOOT_CONFIG="tegra194-mb1-bct-misc-sd-l4t.cfg";
+ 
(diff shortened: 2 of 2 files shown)


danielfullmer:
IIUC, upstream L4T uses TBCDTB_FILE for the device tree intended for the edk2 firmware, and DTB_FILE for the Linux kernel. There can be relevant differences, removing things that aren't necessary for the running Linux system.  Since we don't use the kernel-dtb partitions, we generally don't have this distinction in jetpack-nixos. However, for this one, based on diffing the device trees for Xavier NX, wouldn't this remove access to `mmc0`, (`/sdhci@3460000`), which is needed to flash parts of the platform firmware on Xavier NX?  On Xavier NX, in an A/B slots configuration, parts of the B slot are on the emmc/sdcard. See `flash_l4t_t194_spi_emmc_p3668.xml` for example. WIth your change,  I'm surprised if the initrd flash script would even work, since it would lose access to the emmc/sdcard.

elliotberman:
> However, for this one, based on diffing the device trees for Xavier NX, wouldn't this remove access to mmc0, (/sdhci@3460000), which is needed to flash parts of the platform firmware on Xavier NX?

sdhci@3460000 is the unconnected emmc controller. sdhci@3440000 is the controller for the sdcard. See: https://forums.developer.nvidia.com/t/xavier-nx-cant-find-mmcblk0p1-on-custom-carrier-board/262342/13

> WIth your change, I'm surprised if the initrd flash script would even work, since it would lose access to the emmc/sdcard.

Actually, I think it has not been working properly for some time. The recent changes I made to to ensure the secondary_gpt is placed at the end of the disk uncovered this issue because /dev/mmcblk0 doesn't exist. We've been dd'ing to a tempfile. We've not caught this because we never wipe the sdcard and the expected data persists on the sdcard from legacy flashing script. The sdcard is populating at /dev/mmcblk1 and this change aims to ensure that it ends up as /dev/mmcblk0 as the initrd flash requires.

elliotberman:
> Are you sure we're talking about the same sdcard? On the xavier-nx-devkit it has an SDcard on the SoM module itself, which on xavier-nx-emmc-devkit is replaced with an emmc. These are is separate from the SDcard on the devkit module.

The MMC controller for the xavier-nx-emmc and the xavier-nx-sdcard are different. On xavier-nx-emmc it's the sdhci@3460000. On xavier-nx-sdcard, it's sdhci@3440000.

Here's the output of xavier-nx-devkit (sdcard) with this patch
```
$ ls -l /sys/block/mmcblk0
lrwxrwxrwx 1 root root 0 Jan 19 16:15 mmcblk0 -> ../devices/platform/3400000.sdhci/mmc_host/mmc0/mmc0:59b4/block/mmcblk0
```

The xavier-nx-emmc have different conf files which point to the different XML and DTBs needed to describe the hardware differences.

nvidia/platform/t19x/jakku/kernel-dts/tegra194-p3668-0000-p3509-0000.dts
and nvidia/platform/t19x/jakku/kernel-dts/tegra194-p3668-0001-p3509-0000.dts

modules/flash-script.nix
```diff
@@ -369,6 +369,8 @@ in
 
     hardware.nvidia-jetpack.flashScriptOverrides.fuseArgs = lib.mkAfter [ cfg.flashScriptOverrides.configFileName ];
 
+    hardware.nvidia-jetpack.flashScriptOverrides.patches = lib.mkIf ((cfg.som == "xavier-nx") && (cfg.carrierBoard == "devkit")) [ ../pkgs/r35-bsp.patch ];
```
danielfullmer: I'd recommend moving this to `modules/devices.nix`, since it's configuration dependent on a specific SoM.

Do you mean to just do `cfg.som == "xavier-nx"`, or do you meant to cover `xavier-nx-emmc` as well? (Your patch modifies both)

Also, could you rename the patch to something indicative of what it's doing? There should be a paragraph or so of comment explaining what this is and why we need it. (Because I know at least I'm going to forget in the future :) )

elliotberman:
> It doesn't need to happen in this PR, but I don't think the initrdFlashScript should just assume `mmc0` is the device it should flash platform firmware to. We can know exactly which device type we're on, so shouldn't we just directly write to the block device on the device tree controller we're expecting (for example by `/sys/devices/platform/3400000.sdhci/...` path)?

I don't think we have a way of knowing exactly which block device we should be writing to. At least, I don't think L4T configuration files don't tell us enough to infer it directly. We could add this as a unique feature/requirement for jetpack-nixos.

danielfullmer (approved):

PR #437, opened 2026-01-19
Fixes for GCC 15

danielfullmer:
###### Description of changes

Fixes to display-driver, oot-modules and uefi-firmware for GCC 15. The principal difference causing issues is that GCC 15 defaults to `-std=gnu23`, which leads to various compilation errors.

###### Testing

Built toplevel system closures for the following with `--override-input nixpkgs nixpkgs/master`
- [x] `xavier-agx-devkit`
- [x] `orin-agx-devkit-jp5`
- [x] `orin-agx-devkit`
- [x] `thor-agx-devkit`

Also tested capsule update and booting into `orin-agx-devkit` (JP6)


pkgs/kernels/r35/0001-Update-conftest-with-kernel-KBUILD-options.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/kernel-open/conftest.sh b/kernel-open/conftest.sh
+index 5c00771..fb48398 100755
+--- a/kernel-open/conftest.sh
++++ b/kernel-open/conftest.sh
+@@ -190,7 +190,8 @@ test_headers() {
+ build_cflags() {
+     BASE_CFLAGS="-O2 -D__KERNEL__ \
+ -DKBUILD_BASENAME=\"#conftest$$\" -DKBUILD_MODNAME=\"#conftest$$\" \
+--nostdinc -isystem $ISYSTEM"
++-nostdinc -isystem $ISYSTEM -std=gnu11 -fshort-wchar -funsigned-char \
```
elliotberman, line 19: If we want to be super correct, [5.10 is gnu89
](https://origin.kernel.org/doc/html/v5.10/process/programming-language.html). gnu11 switch happened in [5.18](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=50560ce6a0bdab2fc37384c52aa02c7043909d2c). So JP5 and 6 should use gnu89 and JP7 can use gnu11

```suggestion
+-nostdinc -isystem $ISYSTEM -std=gnu89 -fshort-wchar -funsigned-char \
```

pkgs/kernels/r36/0002-ANDURIL-Add-some-missing-BASE_CFLAGS.patch
```diff
@@ -23,7 +23,7 @@ index 391374f..b89a15e 100644
      BASE_CFLAGS="-O2 -D__KERNEL__ \
  -DKBUILD_BASENAME=\"#conftest$$\" -DKBUILD_MODNAME=\"#conftest$$\" \
 --nostdinc -isystem $ISYSTEM \
-+-nostdinc -isystem $ISYSTEM -fshort-wchar -funsigned-char \
++-nostdinc -isystem $ISYSTEM -std=gnu11 -fshort-wchar -funsigned-char \
```
elliotberman, line 26:
```suggestion
+-nostdinc -isystem $ISYSTEM -std=gnu89 -fshort-wchar -funsigned-char \
```

pkgs/kernels/r36/0003-Fix-conftest-use-with-gcc15.patch
```diff
@@ -0,0 +1,40 @@
...
+diff --git a/scripts/conftest/conftest.sh b/scripts/conftest/conftest.sh
+index 2346a86e..a8428066 100755
+--- a/scripts/conftest/conftest.sh
++++ b/scripts/conftest/conftest.sh
+@@ -90,7 +90,7 @@ build_cflags() {
+     ISYSTEM=`$CC -print-file-name=include 2> /dev/null`
+     BASE_CFLAGS="-O2 -D__KERNEL__ \
+ -DKBUILD_BASENAME=\"#conftest$$\" -DKBUILD_MODNAME=\"#conftest$$\" \
+--nostdinc -isystem $ISYSTEM \
++-nostdinc -isystem $ISYSTEM -std=gnu11 \
```
elliotberman, line 34:
```suggestion
+-nostdinc -isystem $ISYSTEM -std=gnu89 \
```

elliotberman (approved):

pkgs/kernels/r35/0001-Update-conftest-with-kernel-KBUILD-options.patch
danielfullmer: True, but the prior behavior for conftest was using std=gnu11 by virtue of that being gcc's default, and I'd prefer not to change it since it was working.  Plus, if we ever do "bring your own kernel" for JP6, people could be building oot-modules against >=5.18

PR #446, opened 2026-01-27
initrdFlash: Wait for USB device controller to probe

elliotberman:
###### Description of changes

Our script to set up the USB gadget may run before the USB device controller finishes probing. Since we are in simple busybox shell, just spin until it exists.

###### Testing

- [x] initrdFlash orin-agx-devkit

### DIFF
--- a/overlay-with-config.nix
+++ b/overlay-with-config.nix
@@ -133,9 +133,16 @@ final: prev: (
 
               ln -s $gadget/functions/acm.usb0 $gadget/configs/c.1/
 
+              # Disable autosuspend
               if [ -w /sys/bus/usb/devices/usb2/power/control ] ; then
                 echo on >/sys/bus/usb/devices/usb2/power/control
               fi
+
+              # Wait for udc to finish probing, if not done already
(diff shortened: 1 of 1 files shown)


overlay-with-config.nix
```diff
@@ -133,10 +133,14 @@ final: prev: (
 
               ln -s $gadget/functions/acm.usb0 $gadget/configs/c.1/
 
+              # Disable autosuspend
               if [ -w /sys/bus/usb/devices/usb2/power/control ] ; then
                 echo on >/sys/bus/usb/devices/usb2/power/control
               fi
 
+              # Wait for udc to finish probing, if not done already
+              while [[ -z "$(ls /sys/class/udc | head -n 1)" ]] ; do sleep 1 ; done
```
danielfullmer: How long does this typically take to finish probing?
Could you add a message about "Waiting for udc to finish probing..." in the sleep loop?

elliotberman: I suspect it takes on the order of milliseconds, so I don't think the message would be too helpful

PR #451, opened 2026-02-02
vpi-samples: Exclude 11-fisheye when opencv at least 4.10 

elliotberman:
###### Description of changes

The 11-fisheye sample explicitly tests for OpenCV 4.10 and complains
about incompatability.

/build/source/opt/nvidia/vpi4/samples/11-fisheye/main.cpp:39:6: error: #error "OpenCV >= 4.10 isn't supported"
  39 | #    error "OpenCV >= 4.10 isn't supported"

###### Testing

- [x] build/run vpi-test

### DIFF
--- a/pkgs/samples/vpi-samples.nix
+++ b/pkgs/samples/vpi-samples.nix
@@ -32,11 +32,17 @@ backendStdenv.mkDerivation {
   buildInputs = [ cuda_cudart opencv vpi ]
     ++ lib.optionals (l4tAtLeast "38") [ l4t-pva ];
 
+  # Sample directories which we won't build.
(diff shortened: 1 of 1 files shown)


danielfullmer:
Could you actually just block the specific sample that doesn't build with opencv? We may want to run other samples in the future.

pkgs/samples/vpi-samples.nix
```diff
@@ -17,6 +17,8 @@ let
     ;
 
   vpiMajor = lib.versions.major vpi.version;
+
```
danielfullmer: Can you remove extraneous whitespace change here

PR #455, opened 2026-02-05
nixos: graphics: Add l4t-video-codec-openrm to hardware.graphics.extraPackages

elliotberman:
###### Description of changes

Provides libs necessary for encoding and decoding video.

###### Testing

- [x] Run multimedia samples

### DIFF
--- a/modules/graphics.nix
+++ b/modules/graphics.nix
@@ -67,6 +67,7 @@ in
                 "l4t-nvml" # JP6+
                 "l4t-nvsci"
                 "l4t-pva"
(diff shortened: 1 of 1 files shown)


danielfullmer:
Is this still needed? What is a concrete example of something that fails if we don't have this? Is there a multimedia-samples test that covers this, and if not, can we add a small test?

E-Gruda:
> Is this still needed? What is a concrete example of something that fails if we don't have this? Is there a multimedia-samples test that covers this, and if not, can we add a small test?

without this change multimedia is broken on JP7

PR #458, opened 2026-02-07
Rework L4T container file generation

danielfullmer:
###### Description of changes

Simplify the generation script by pulling csv from `nvidia-l4t-init`. Removed extraneous minor version number from json file name.  (We only ever have one minor version of jetpack at a time, and having them use the same filename makes diffs easier).
Regenerated the `r35-l4t.json` file, which was somehow out-of-date.

###### Testing

Ran our OCI test on an Orin AGX devkit.

### DIFF
--- a/pkgs/containers/r35-l4t.json
+++ b/pkgs/containers/r35-l4t.json
@@ -1,14 +1,14 @@
 {
   "nvidia-l4t-3d-core": [
     "/usr/lib/aarch64-linux-gnu/tegra/libGLX_nvidia.so.0",
-    "/usr/lib/aarch64-linux-gnu/tegra/libnvidia-eglcore.so.35.6.1",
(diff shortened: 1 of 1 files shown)


pkgs/containers/genL4tJson.nix
```diff
@@ -0,0 +1,8 @@
+{ runCommand, buildPackages, unpackedDebsFilenames, l4tAtLeast }:
+
+let
+  l4tCsvFilename = if l4tAtLeast "36" then "drivers.csv" else "l4t.csv";
+in
+runCommand "l4t.json" { nativeBuildInputs = [ buildPackages.python3 buildPackages.dpkg ]; } ''
```
ConnorBaker: Is `buildPackages` necessary? Both should come from the top-level scope, which is spliced.

danielfullmer: Nope, you're right. Will fix.

PR #461, opened 2026-02-12
nvpmodel: dynamically select conf file at runtime

sm1ll3r:
###### Description of changes

Adds a small setup script that symlinks the matching conf file for a given som.
Based on nvidia-l4t-init/etc/systemd/nvpower.sh but better!

This also fixes #168

###### Testing

Checked that nvpmodel loads successfully and points to the correct nvpmodel conf file on:

- [x] thor-agx
- [x] xavier-agx-industrial
- [x] xavier-nx
- [x] xavier-agx
- [x] orin-agx jp5 / jp6
- [x] orin-agx-industrial jp5 / jp6
- [x] orin-nx-16gig jp5 / jp6
- [x] orin-nx-8gig jp5 / jp6
- [x] orin-nano jp5 / jp6

### DIFF
--- a/modules/devices.nix
+++ b/modules/devices.nix
@@ -9,19 +9,6 @@ let
 
   cfg = config.hardware.nvidia-jetpack;
 
-  # TODO: Replace default selection with dynamic logic from nvidia-l4t-init/etc/systemd/nvpower.sh
-  nvpModelConf = {
-    orin-agx = "${pkgs.nvidia-jetpack.l4t-nvpmodel}/etc/nvpmodel/nvpmodel_p3701_0000.conf";
-    orin-agx-industrial = "${pkgs.nvidia-jetpack.l4t-nvpmodel}/etc/nvpmodel/nvpmodel_p3701_0008.conf";
-    orin-nx = "${pkgs.nvidia-jetpack.l4t-nvpmodel}/etc/nvpmodel/nvpmodel_p3767_0000${lib.optionalString cfg.super "_super"}.conf";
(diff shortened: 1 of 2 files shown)


modules/nvpmodel.nix
```diff
@@ -36,12 +58,12 @@ in
         Type = "oneshot";
         Restart = "on-failure";
         RestartSec = "2s";
-        ExecStart = "${pkgs.nvidia-jetpack.l4t-nvpmodel}/bin/nvpmodel -f ${cfg.configFile}" + lib.optionalString (cfg.profileNumber != null) " -m ${builtins.toString cfg.profileNumber}";
+        ExecStartPre = "${setupConf}/bin/setupConf";
```
elliotberman:
```suggestion
        ExecStartPre = lib.getExe setupConf;
```

modules/devices.nix
```diff
@@ -56,8 +43,7 @@ in
 {
   config = lib.mkIf cfg.enable (lib.mkMerge [{
     # Turn on nvpmodel if we have a config for it.
-    services.nvpmodel.enable = mkIf (nvpModelConf ? "${cfg.som}") (mkDefault true);
-    services.nvpmodel.configFile = mkIf (nvpModelConf ? "${cfg.som}") (mkDefault nvpModelConf.${cfg.som});
+    services.nvpmodel.enable = mkDefault true;
```
sm1ll3r, line 46: Using this for testing but not sure what this should be set to now since we determine the conf at run time

elliotberman: mkdefault true makes sense to me

danielfullmer (approved):
Will merge this as-is, but we should consider just running nvpower.sh (perhaps with modifications), since it has a bunch of conditional logic to handle emulated variants, igx, etc.

PR #486, opened 2026-03-27
Add NixOS option to configure UEFI boot menu timeout via EFI variable

Princemachiavelli:
###### Description of changes
Add hardware.nvidia-jetpack.firmware.uefi.bootMenuTimeout option (0-65535 seconds) that writes the UEFI Timeout global variable at boot via a systemd oneshot service.

###### Testing

Successfully tested setting values of 0 and 25 on orin-agx-devkit.

### DIFF
--- a/modules/boot-menu-timeout.nix
+++ b/modules/boot-menu-timeout.nix
@@ -1,0 +1,60 @@
+{ config, lib, pkgs, ... }:
+
+let
+  inherit (lib)
+    mkIf
+    mkOption
+    types;
+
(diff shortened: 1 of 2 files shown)


Princemachiavelli:
> I would, however, be open to a setting in the uefi firmware build that changes this default.

Sounds like a good plan. I wonder how UEFI capsule updates will work, will the previous "default" value of the `Timeout` UEFI variable persist across capsule updates such that runtime changes are the only way to update the value?

PR #487, opened 2026-03-30
Initial options for "bring your own kernel"

danielfullmer:
###### Description of changes

This PR adds a `hardware.nvidia-jetpack.kernel.useVendorProvided` option that defaults to true, with the aim to allow users to more easily "bring their own kernel".  Users could have done this themselves in the past by overriding `boot.kernelPackages` with a higher priority setting, but adding this option makes it slightly easier to use and allows us to easily condition other settings on whether we're using the vendor kernel or not.

Additionally, this applies the additional out-of-tree kernel modules via via the [kernelPackageExtensions](https://github.com/NixOS/nixpkgs/pull/448069) option introduced in NixOS 25.11. Finally, we move most kernel-related configs/options into a new file for better separation.

###### Testing

In progress

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,21 +260,9 @@ from Jetson Linux.
 
 The NixOS module uses these sets by default.
 
-On JetPack 6+, however, you may use a mainline package set instead.
+On JetPack 6+, however, you may use a mainline kernel instead.
+You can disable Nvidia's provided vendor kernel by setting `hardware.nvidia-jetpack.kernel.useVendorProvided = false;`.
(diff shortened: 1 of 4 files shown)


TanelDettenborn:
Could this PR include options for Nvidia OOT patching? I think it would significantly improve support for BYOK. Depends on kernel/environment, but you might need to patch NVIDIA-OOT with "compatibility" patches. Currently it is not doable - easily at least.

For example options for these: https://github.com/anduril/jetpack-nixos/blob/master/pkgs/kernels/r36/oot-modules.nix#L22-L53

PR #489, opened 2026-04-07
jetpackDowngrade: allow configurable downgrade levels

Princemachiavelli:
###### Description of changes

In some cases, it may be desirable or logistically required to allow downgrading the L4T userspace components without reflashing the device (such as for rapid local development).  This PR makes the downgrade prevention check configurable against the SemVer level of the L4T version. Users can optionally allow "minor" or "patch" level downgrades which, in most cases, won't cause regressions while still blocking "major" level downgrades which are known to soft brick devices.

###### Testing
Manually tested each downgrade level on orin-agx-devkit.

### DIFF
--- a/modules/pre-switch-checks.nix
+++ b/modules/pre-switch-checks.nix
@@ -1,47 +1,92 @@
-{ pkgs, ... }:
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.hardware.nvidia-jetpack;
+in
 {
+  options.hardware.nvidia-jetpack.firmware.maxAllowedDowngrade = lib.mkOption {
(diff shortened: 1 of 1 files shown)


Princemachiavelli:
Example output:
```
[root@nixos:~]# ./pre-switch-checks /run/current-system boot
Error: L4T version downgrade detected!
Running L4T version: 36.5.0
Target L4T version:  36.4.9
Downgrade protection level: patch
Downgrading the L4T version is not supported.
This requires re-flashing the device firmware.
Pre-switch check 'jetpackDowngrade' failed
```
Check no longer outputs anything unless the downgrade is blocked or we fail to read the `bios_version` file.

elliotberman (approved):

PR #496, opened 2026-04-30
feat(optee): add support for fTPM on JetPack 6

zengraf:
###### Description of changes

Rebases PR #484 (fTPM support) on top of #480 (optee-nix refactor), and addresses all review comments from the original PR.

Adds `hardware.nvidia-jetpack.firmware.optee.ftpm` options that build the MS TPM 2.0 reference TA + the fTPM helper TA, embed them as early TAs in `tos.img`, and load `tpm_ftpm_tee` after `tee-supplicant` (now `Type=notify`).

Currently scoped to l4t r36.x: r35 lacks fTPM source; r38 refactored fTPM compilation around `CFG_MS_TPM_20_REF` and isn't wired up here yet.

**Changes from PR #484 (review comments)**

- Use default GCC with `NIX_CFLAGS_COMPILE` instead of pinning `gcc13` for the MS TPM 2.0 ref TA build.
- Renamed `unsafeInjectEPS` → `unsecureInjectEPS`. Made the EPS value configurable (`unsecureInjectEPS.value`); falls back to a random, persisted-in-`/var/lib` EPS when null.
- New `firmware.optee.ftpm.measuredBoot` option, gating `CFG_TA_MEASURED_BOOT` (in the TA build) and `CFG_CORE_TPM_EVENT_LOG` (in OP-TEE OS) together.
- New `firmware.optee.ftpm.taLogLevel` option (default 0, NVIDIA's default), separate from the global `taLogLevel`.
- `ftpm-driver.service` now `After=local-fs.target systemd-modules-load.service` in addition to `tee-supplicant.service`.
- Assertion blocks `ftpm.enable` on l4t r35 / r38.
- The systemd `sd_notify` patch is applied via `fetchpatch` from upstream `OP-TEE/optee_client@a5b1ffcd`, flanked by two local patches to handle `nvme_rpmb.c` present in nv-optee but not upstream.

**Changes from #480 refactor**

- New `pkgs/optee/{ftpmHelperTa,msTpm20RefTa}.nix` follow the per-derivation pattern.
- `pkgs/optee/optee-os.nix` gains overrideable fTPM attrs (`enableFTPM`, `measuredBoot`, `unsecureInjectEPS`, `ftpmHelperTa`, `msTpm20RefTa`); `overlay-with-config.nix` sets them from the NixOS config.
- `taDevKit` unconditionally clears `uefi-firmware`, `ftpmHelperTa`, and `msTpm20RefTa` to break an `optee-os → fTPM TA → taDevKit → optee-os` evaluation cycle and prevent a duplicate EDK2 build when fTPM is enabled.

###### Testing

- Tested with ftpm enabled and disabled on Jetson Orin Nano with Jetpack 6. The functionality is verified with `tpm2-tools`.

### DIFF
--- a/modules/optee.nix
+++ b/modules/optee.nix
@@ -11,6 +11,8 @@ let
 
   cfg = config.hardware.nvidia-jetpack.firmware.optee;
 
+  inherit (pkgs.nvidia-jetpack) l4tAtLeast l4tOlder;
+
 in
 {
   imports = [
@@ -68,6 +70,63 @@ in
         '';
       };
 
+      ftpm = {
+        enable = mkOption {
+          type = types.bool;
+          default = false;
+          description = ''
+            Enable fTPM (Firmware TPM) support. Builds the MS TPM 2.0
+            reference TA and the fTPM helper TA, embeds them in the OP-TEE
+            tos.img firmware, and loads the tpm_ftpm_tee kernel driver
+            after tee-supplicant is ready. Toggling this option requires
+            re-flashing the platform firmware.
+
+            Currently only supported on l4t r36.x. r35 lacks fTPM source;
+            r38 refactored fTPM compilation (see CFG_MS_TPM_20_REF) and is
+            not yet wired up here.
+          '';
+        };
+
+        measuredBoot = mkOption {
+          type = types.bool;
+          default = false;
+          description = ''
+            Enable measured-boot support. Sets CFG_TA_MEASURED_BOOT in the
+            MS TPM 2.0 reference TA build and CFG_CORE_TPM_EVENT_LOG in
+            OP-TEE OS.
+          '';
+        };
+
+        taLogLevel = mkOption {
+          type = types.int;
+          default = 0;
+          description = ''
+            CFG_TA_LOG_LEVEL passed to the MS TPM 2.0 reference TA build.
+            Independent from
+            hardware.nvidia-jetpack.firmware.optee.taLogLevel.
+          '';
+        };
+
+        unsecureInjectEPS = {
+          enable = mkEnableOption ''
+            unsecure EPS injection. The fTPM TA exposes functionality to
+            inject a custom Endorsement Primary Seed (EPS). This is
+            effectively a TPM backdoor and should ONLY be used for testing
+          '';
+
+          value = mkOption {
+            type = types.nullOr types.str;
+            default = null;
+            example = "0xdeadbeef...";
+            description = ''
+              Specific EPS value (64-byte hex string with 0x prefix) to
+              inject. If null, a random EPS is generated on first boot
+              and persisted to /var/lib/unsecureInjectEPS.hex.
+            '';
+          };
+        };
+      };
+
       patches = mkOption {
         type = types.listOf types.path;
         default = [ ];
@@ -108,43 +167,131 @@ in
     };
   };
 
-  config = mkIf config.hardware.nvidia-jetpack.enable {
-    hardware.nvidia-jetpack.firmware.optee.supplicant.trustedApplications =
-      lib.optional cfg.pkcs11Support pkgs.nvidia-jetpack.pkcs11Ta
-      ++ lib.optional cfg.xtest pkgs.nvidia-jetpack.opteeXtest.tas;
+  config = mkIf config.hardware.nvidia-jetpack.enable (lib.mkMerge [
+    {
+      assertions = [{
+        # TODO: extend to r38 once fTPM is wired up via CFG_MS_TPM_20_REF.
+        assertion = !cfg.ftpm.enable || (l4tAtLeast "36" && l4tOlder "38");
+        message = ''
+          hardware.nvidia-jetpack.firmware.optee.ftpm.enable currently
+          requires l4t r36.x. r35 lacks fTPM source; r38 refactored fTPM
+          compilation to use the CFG_MS_TPM_20_REF flag and is not yet
+          supported here.
+        '';
+      }];
 
-    hardware.nvidia-jetpack.firmware.optee.supplicant.plugins =
-      lib.optional cfg.xtest pkgs.nvidia-jetpack.opteeXtest.plugins;
+      hardware.nvidia-jetpack.firmware.optee.supplicant.trustedApplications =
+        lib.optional cfg.pkcs11Support pkgs.nvidia-jetpack.pkcs11Ta
+        ++ lib.optional cfg.xtest pkgs.nvidia-jetpack.opteeXtest.tas;
 
-    systemd.services.tee-supplicant =
-      let
-        teeApplications = pkgs.symlinkJoin {
-          name = "tee-applications";
-          paths = cfg.supplicant.trustedApplications;
+      hardware.nvidia-jetpack.firmware.optee.supplicant.plugins =
+        lib.optional cfg.xtest pkgs.nvidia-jetpack.opteeXtest.plugins;
+
+      systemd.services.tee-supplicant =
+        let
+          teeApplications = pkgs.symlinkJoin {
+            name = "tee-applications";
+            paths = cfg.supplicant.trustedApplications;
+          };
+
+          supplicantPlugins = pkgs.symlinkJoin {
+            name = "tee-supplicant-plugins";
+            paths = cfg.supplicant.plugins;
+          };
+
+          args = lib.escapeShellArgs (
+            [
+              "--ta-path=${teeApplications}"
+              "--plugin-path=${supplicantPlugins}"
+            ]
+            ++ cfg.supplicant.extraArgs
+          );
+        in
+        mkIf cfg.supplicant.enable {
+          description = "Userspace supplicant for OPTEE-OS";
+          serviceConfig = {
+            # tee-supplicant is patched to call sd_notify(READY=1); without
+            # NOTIFY_SOCKET this is a harmless no-op for non-systemd users.
+            Type = "notify";
+            ExecStart = "${pkgs.nvidia-jetpack.opteeClient}/bin/tee-supplicant ${args}";
+            Restart = "always";
+          };
+          wantedBy = [ "multi-user.target" ];
         };
 
-        supplicantPlugins = pkgs.symlinkJoin {
-          name = "tee-supplicant-plugins";
-          paths = cfg.supplicant.plugins;
-        };
+      environment.systemPackages = lib.optional cfg.xtest pkgs.nvidia-jetpack.opteeXtest;
+    }
 
-        args = lib.escapeShellArgs (
-          [
-            "--ta-path=${teeApplications}"
-            "--plugin-path=${supplicantPlugins}"
-          ]
-          ++ cfg.supplicant.extraArgs
-        );
-      in
-      mkIf cfg.supplicant.enable {
-        description = "Userspace supplicant for OPTEE-OS";
-        serviceConfig = {
-          ExecStart = "${pkgs.nvidia-jetpack.opteeClient}/bin/tee-supplicant ${args}";
-          Restart = "always";
-        };
-        wantedBy = [ "multi-user.target" ];
-      };
+    (mkIf cfg.ftpm.enable {
+      boot.kernelModules = [ "tpm" ];
+      boot.initrd.availableKernelModules = [ "tpm" "tpm_ftpm_tee" ];
+      # tpm_ftpm_tee must load after tee-supplicant is up.
+      boot.blacklistedKernelModules = [ "tpm_ftpm_tee" ];
 
-    environment.systemPackages = lib.optional cfg.xtest pkgs.nvidia-jetpack.opteeXtest;
-  };
+      boot.kernelPatches = [{
+        name = "fTPM_tee";
+        patch = null;
+        structuredExtraConfig = with lib.kernel; {
+          TCG_TPM = module;
+          TCG_FTPM_TEE = module;
+        };
+        features.fTPM_tee = true;
+      }];
+
+      systemd.services.ftpm-driver =
+        let
+          epsFile = "/var/lib/unsecureInjectEPS.hex";
+          helper = "${pkgs.nvidia-jetpack.ftpmHelperTa}/bin/nvftpm-helper-app";
+
+          epsInjectScript =
+            if cfg.ftpm.unsecureInjectEPS.value != null then ''
+              echo "Injecting configured EPS into fTPM..."
+              ${helper} -g ${lib.escapeShellArg cfg.ftpm.unsecureInjectEPS.value}
+            '' else ''
+              EPS_FILE=${lib.escapeShellArg epsFile}
+              EPS_SIZE=64
+
+              if [ ! -f "$EPS_FILE" ]; then
+                echo "Generating random EPS for fTPM..."
+                mkdir -p "$(dirname "$EPS_FILE")"
+                EPS_HEX="0x$(${pkgs.coreutils}/bin/dd if=/dev/urandom bs=1 count=$EPS_SIZE 2>/dev/null | ${pkgs.xxd}/bin/xxd -p -c 256)"
+                echo "$EPS_HEX" > "$EPS_FILE"
+                chmod 600 "$EPS_FILE"
+                echo "EPS saved to $EPS_FILE"
+              else
+                echo "Using existing EPS from $EPS_FILE"
+              fi
+
+              EPS_VALUE=$(cat "$EPS_FILE")
+              echo "Injecting EPS into fTPM..."
+              ${helper} -g "$EPS_VALUE"
+            '';
+
+          script = pkgs.writeShellScript "ftpm-driver-load" ''
+            set -euo pipefail
+
+            ${lib.optionalString cfg.ftpm.unsecureInjectEPS.enable epsInjectScript}
+
+            echo "Loading tpm_ftpm_tee driver..."
+            ${pkgs.kmod}/bin/modprobe tpm_ftpm_tee
+          '';
+        in
+        {
+          description = "Load fTPM driver after TEE supplicant";
+          after = [
+            "tee-supplicant.service"
+            "local-fs.target"
+            "systemd-modules-load.service"
+          ];
+          requires = [ "tee-supplicant.service" ];
+          before = [ "tpm2.target" ];
+          wantedBy = [ "multi-user.target" ];
+          serviceConfig = {
+            Type = "oneshot";
+            RemainAfterExit = true;
+            ExecStart = script;
+          };
+        };
+    })
+  ]);
 }
--- a/modules/optee.nix
+++ b/modules/optee.nix
@@ -284,7 +284,6 @@ in
             "systemd-modules-load.service"
           ];
           requires = [ "tee-supplicant.service" ];
-          before = [ "tpm2.target" ];
           wantedBy = [ "multi-user.target" ];
           serviceConfig = {
             Type = "oneshot";
--- a/modules/optee.nix
+++ b/modules/optee.nix
@@ -121,7 +121,7 @@ in
             description = ''
               Specific EPS value (64-byte hex string with 0x prefix) to
               inject. If null, a random EPS is generated on first boot
-              and persisted to /var/lib/unsecureInjectEPS.hex.
+              and persisted to /var/lib/optee/ftpm/unsecureInjectEPS.hex.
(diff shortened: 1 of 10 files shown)


zengraf:
Cc @dperezzoghbi, @TanelDettenborn, @danielfullmer

overlay-with-config.nix
```diff
@@ -76,11 +76,42 @@ final: prev: (
...
+          ]
+          ++ lib.optional cfg.firmware.optee.ftpm.measuredBoot
+            "CFG_CORE_TPM_EVENT_LOG=y"
+          ++ lib.optional cfg.firmware.optee.ftpm.unsecureInjectEPS.enable
+            (lib.warn
+              "fTPM is using UNSECURE Endorsement Primary Seed (EPS) injection."
+              "CFG_JETSON_FTPM_HELPER_INJECT_EPS=y");
+        earlyTaPaths = prevAttrs.earlyTaPaths or [ ]
+          ++ lib.optionals
+            (cfg.firmware.optee.ftpm.enable && finalJetpack.socType == "t234") [
```
dperezzoghbi: I dont think we need the socType check here. It should only be conditional on optee.ftpm enablement.

This check on socType should be elsewhere, perhaps with an assertion that socType == t234. Otherwise, later we'll miss this when we support other soc types.

modules/optee.nix
```diff
@@ -108,43 +167,131 @@ in
...
-        wantedBy = [ "multi-user.target" ];
-      };
+        features.fTPM_tee = true;
+      }];
 
-    environment.systemPackages = lib.optional cfg.xtest pkgs.nvidia-jetpack.opteeXtest;
-  };
+      systemd.services.ftpm-driver =
+        let
+          epsFile = "/var/lib/unsecureInjectEPS.hex";
```
dperezzoghbi: Leaving this in the root /var/lib is probably bad practice. Lets change this to be `/var/lib/optee/ftpm/unsecureInjectEPS.hex`

Also set `StateDirectory = "optee/ftpm"` on the ftpm-driver service unit.

modules/optee.nix
```diff
@@ -108,43 +167,131 @@ in
...
+          helper = "${pkgs.nvidia-jetpack.ftpmHelperTa}/bin/nvftpm-helper-app";
+
+          epsInjectScript =
+            if cfg.ftpm.unsecureInjectEPS.value != null then ''
+              echo "Injecting configured EPS into fTPM..."
+              ${helper} -g ${lib.escapeShellArg cfg.ftpm.unsecureInjectEPS.value}
```
dperezzoghbi: Let's write it to the same file instead. This way there's a "standard" method of verifying that the EPS is insecure and being able to see the raw EPS easily.

```nix
epsInjectScript = ''
  EPS_FILE=${lib.escapeShellArg epsFile}
  EPS_SIZE=64
  if [ ! -f "$EPS_FILE" ]; then
    ${if cfg.ftpm.unsecureInjectEPS.value != null then ''
      echo "Generating configured EPS for fTPM..."
      EPS_HEX="${lib.escapeShellArg cfg.ftpm.unsecureInjectEPS.value}"
    '' else ''
      echo "Generating random EPS for fTPM..."
      mkdir -p "$(dirname "$EPS_FILE")"
      EPS_HEX="0x$(${pkgs.coreutils}/bin/dd if=/dev/urandom bs=1 count=$EPS_SIZE 2>/dev/null | ${pkgs.xxd}/bin/xxd -p -c 256)"
    ''}
    echo "$EPS_HEX" > "$EPS_FILE"
    chmod 600 "$EPS_FILE"
    echo "EPS saved to $EPS_FILE"
  else
    echo "Using existing EPS from $EPS_FILE"
  fi
  EPS_VALUE=$(cat "$EPS_FILE")
  echo "Injecting EPS into fTPM..."
  ${helper} -g "$EPS_VALUE"
'';
```

overlay-with-config.nix
```diff
@@ -76,11 +76,26 @@ final: prev: (
...
+
       optee-os = prevJetpack.optee-os.overrideAttrs (prevAttrs: {
         inherit (finalJetpack) socType;
         inherit (cfg.firmware.optee) taPublicKeyFile coreLogLevel taLogLevel;
         patches = prevAttrs.patches or [ ] ++ cfg.firmware.optee.patches;
         makeFlags = prevAttrs.makeFlags or [ ] ++ cfg.firmware.optee.extraMakeFlags;
+        enableFTPM = cfg.firmware.optee.ftpm.enable;
+        measuredBoot = cfg.firmware.optee.ftpm.measuredBoot;
+        unsecureInjectEPS = cfg.firmware.optee.ftpm.unsecureInjectEPS.enable;
+        ftpmHelperTa = if (cfg.firmware.optee.ftpm.enable && finalJetpack.socType == "t234")
```
dperezzoghbi: I would suggest removing the socType check. Leave those checks in the module's assertions. It wont scale well to put these checks for every chipset. We can just use the assertion in the module.

pkgs/optee/optee-os.nix
```diff
@@ -31,6 +30,18 @@ in
   taLogLevel = finalAttrs.coreLogLevel;
   taPublicKeyFile = null;
 
+  # fTPM — set ftpmHelperTa/msTpm20RefTa to the TA derivations to embed them as early TAs
+  enableFTPM = false;
+  measuredBoot = false;
+  unsecureInjectEPS = false;
+  ftpmHelperTa = null;
+  msTpm20RefTa = null;
+
```
dperezzoghbi, line 39: Maybe we should leave a comment as to why we use the `.stripped.elf` instead of the `.ta`

```nix
# Stripped ELFs are used here. The signed .ta is unnecessary because early
# TAs do not need to be signed, they are embedded directly in the OP-TEE
# binary image.
# See: https://optee.readthedocs.io/en/latest/building/trusted_applications.html#signing-of-tas
# Only REE (file-system-loaded) TAs require signing. Embedded TAs inherit
# the same trust level as OP-TEE OS itself.
```

pkgs/optee/optee-os.nix
```diff
@@ -31,6 +30,18 @@ in
...
+  earlyTaPaths = lib.optionals (finalAttrs.ftpmHelperTa != null) [
```
dperezzoghbi: Use `lib.optionals finalAttrs.enableFTPM` rather than
`lib.optionals (finalAttrs.ftpmHelperTa != null)`. This way if
`enableFTPM = true` but the TA paths weren't provided, you get an
immediate eval error rather than a silent disable of the fTPM.

pkgs/optee/optee-os.nix
```diff
@@ -70,6 +81,15 @@ in
     "CFG_STMM_PATH=${uefi-firmware}/standalonemm_optee.bin"
   ])
   ++ (lib.optional (finalAttrs.taPublicKeyFile != null) "TA_PUBLIC_KEY=${finalAttrs.taPublicKeyFile}")
+  ++ lib.optionals finalAttrs.enableFTPM [
```
dperezzoghbi: The `enableFTPM` condition is currently duplicated across three
`lib.optional(s)` calls. Wrap them in a single outer
`lib.optionals finalAttrs.enableFTPM (...)` so sub-feature flags only
need their own condition:

```nix
++ lib.optionals finalAttrs.enableFTPM ([
  "CFG_REE_STATE=y"
  "CFG_JETSON_FTPM_HELPER_PTA=y"
]
++ lib.optional finalAttrs.measuredBoot "CFG_CORE_TPM_EVENT_LOG=y"
++ lib.optional finalAttrs.unsecureInjectEPS
  (lib.warn
    "fTPM is using UNSECURE Endorsement Primary Seed (EPS) injection."
    "CFG_JETSON_FTPM_HELPER_INJECT_EPS=y"))
```

This makes the hierarchy explicit and avoids repeating the same guard.

modules/optee.nix
```diff
@@ -108,43 +167,131 @@ in
...
+                echo "$EPS_HEX" > "$EPS_FILE"
+                chmod 600 "$EPS_FILE"
+                echo "EPS saved to $EPS_FILE"
+              else
+                echo "Using existing EPS from $EPS_FILE"
+              fi
+
+              EPS_VALUE=$(cat "$EPS_FILE")
+              echo "Injecting EPS into fTPM..."
+              ${helper} -g "$EPS_VALUE"
```
dperezzoghbi: NVIDIA changed the `nvftpm-helper-app` CLI flags between JetPack
releases. On JP5 (r35.5.0+), `-g` injects the EPS. On JP6+ (r36+),
new options were added and the flags shifted — `-g` became "query ECID"
and `-m` is now the EPS injection flag:

```c
// JP5 (r35.5.0–r35.6.x)
{0, 'g', "EPS", 0, "Inject an EPS(...) into fTPM"},

// JP6+ (r36+)
{0, 'g', NULL, 0, "Query the device ECID value"},
{0, 'm', "EPS", 0, "Inject an EPS(...) into fTPM."},
```

The original code hardcoded `-g`, which is correct for JP5 but wrong
for JP6+. Since this PR adds r35 support alongside r36, the flag needs
to be version-gated.

`epsFlag = if l4tAtLeast "36" then "-m" else "-g";`
`${helper} ${epsFlag} "$EPS_VALUE"`

I still plan on adding JP5 support. I can add this, switching if you want, but at least switch to `-m` for this jp6 feature.

zengraf:
Cc @dperezzoghbi, @TanelDettenborn, @danielfullmer, @elliotberman

pkgs/optee/opteeClient.nix
```diff
@@ -21,6 +21,16 @@ stdenv.mkDerivation {
...
+    ]) ++ [
+      ./0002-tee-supplicant-Makefile-drop-nvme-rpmb.patch
+      (fetchpatch {
+        name = "tee-supplicant-add-systemd-sd_notify-support.patch";
+        url = "https://github.com/OP-TEE/optee_client/commit/a5b1ffcd26e328af0bbf18ab448a38ecd558e05c.patch";
+        stripLen = 1;
+        extraPrefix = "optee/optee_client/";
+        hash = "sha256-QDE6wKxA3kvLfcb5ILZZqgJ7WC3UGFmuKtrRP7MwPfM=";
+      })
+      ./0003-tee-supplicant-Makefile-restore-nvme-rpmb.patch
```
TanelDettenborn: Sorry for bundling multiple points into one comment.

- These lines currently break r38 compilation. I believe the Jetson r38 release already includes the `a5b1ffcd26e328af0bbf18ab448a38ecd558e05c` commit, which likely explains the issue.

- I don’t have a strong opinion on whether we should use the drop–fetchpatch–restore approach, carry a rebased patch in the project, or something else. However, I’d really appreciate some comments explaining the rationale behind the chosen patching approach, as that context would make future maintenance much easier.

TanelDettenborn (commented):
I’ve only had a chance to quickly skim through the changes so far and still need to review it in more depth, but one thing I noticed is that this PR currently breaks the r38 build, which will need to be addressed.

Great effort on this - I’d really like to see fTPM support make it into the jetpack-nixos mainline.

dperezzoghbi (changes requested):
Hi @zengraf,
Thanks for cleaning this up — the fetchpatch for sd_notify, the gcc13
removal, and the richer module options are all solid improvements over my
earlier branch (`dperezzoghbi/fTPM-support-3-20`).

Overall this is good, please do check on a few of these requests, especially the `-g` -> `-m` flag.

zengraf:
Thank you for such high quality reviews! I've implemented your suggestions

zacharyweiss:
@zengraf hopefully not to overload this PR: made the minor edits to extend support to Jetpack 7. Tested quickly on a Thor devkit and all seems happy. Cf. https://github.com/zengraf/jetpack-nixos/pull/1 if you'd like to include in this PR, else can wait for this to merge and open against the main repo.

pkgs/optee/taDevKit.nix
```diff
@@ -1,5 +1,8 @@
 { optee-os }:
 optee-os.overrideAttrs (finalAttrs: {
   pname = "optee-ta-dev-kit";
+  uefi-firmware = null;
```
dperezzoghbi, line 4: also need to set enableFTPM = false to avoid ftpmHelperTa = null being coerced into a string (see other comment).

pkgs/optee/optee-os.nix
```diff
@@ -31,6 +30,27 @@ in
...
+  # Stripped ELFs are used here rather than signed .ta files. Early TAs
+  # are embedded directly in the OP-TEE binary image and do not require
+  # signing — they inherit the same trust level as OP-TEE OS itself.
+  # Only REE (file-system-loaded) TAs need to be signed.
+  # See: https://optee.readthedocs.io/en/latest/building/trusted_applications.html#signing-of-tas
+  #
+  # Using enableFTPM (rather than ftpmHelperTa != null) ensures that if
+  # enableFTPM = true but the TA paths weren't provided, you get an
+  # immediate eval error instead of a silent disable of fTPM.
+  earlyTaPaths = lib.optionals finalAttrs.enableFTPM [
```
dperezzoghbi, line 49: when `finalAttrs.ftpmHelperTa` is `null` and `enableFTPM` is true (aka when building `taDevKit` on a system with `ftpm.enable=true`), we get an evaluation error where `finalAttrs.ftpmHelperTa` attempts to be coerced into a string.

dperezzoghbi (commented):
@zengraf Looks good except for one evaluation issue shown below. After this I'm happy to approve.

I have tested on jp5 (35.6.4), jp6 (36.5.0) with enable ftpm on unfused devices using `unsafeInjectEps = true`. I have also built jp5-7 and booted successfully on those.

I also pulled in @zacharyweiss 's patch and tested on jp5-7 which all seem to pass my tests for sealing/unsealing. His seems to address the evaluation issue above. So you can ignore my other comments if you pull this in.

TanelDettenborn:
Quickly tested this PR on Orin NX + Xavier NX devkit alongside @zacharyweiss's [PR](https://github.com/zengraf/jetpack-nixos/pull/1) with iso_minimal. Everything looks good, but I noticed that the fTPM TA panics during system shutdown/teardown. This seems to happen because the fTPM TA requires storage access after tee-supplicant has already stopped.

I have no intention to interfere with the implementation, but I would really like to get this PR merged into master. Therefore, I have opened a separate [PR](https://github.com/zengraf/jetpack-nixos/pull/2) to address the fTPM TA panic. There might be a better solution, but it at least demonstrates the problem and a potential fix.

Assuming @zacharyweiss's PR is accepted and fTPM TA shutdown panic is addressed, this PR looks good. The only other thing I would consider is squashing the commits. Depending on the open PR, this could probably be squashed down to 1-3 commits.

Good work!

modules/optee.nix
```diff
@@ -68,6 +70,63 @@ in
...
+        };
+
+        taLogLevel = mkOption {
+          type = types.int;
+          default = 0;
+          description = ''
+            CFG_TA_LOG_LEVEL passed to the MS TPM 2.0 reference TA build.
+            Independent from
+            hardware.nvidia-jetpack.firmware.optee.taLogLevel.
+          '';
```
elliotberman, line 107: Default to `config.hardware.nvidia-jetpack.optee.taLogLevel`?

pkgs/optee/optee-os.nix
```diff
@@ -31,6 +30,27 @@ in
   taLogLevel = finalAttrs.coreLogLevel;
   taPublicKeyFile = null;
 
+  # fTPM — set ftpmHelperTa/msTpm20RefTa to the TA derivations to embed them as early TAs
+  enableFTPM = false;
+  measuredBoot = false;
+  unsecureInjectEPS = false;
+  ftpmHelperTa = null;
+  msTpm20RefTa = null;
```
elliotberman, line 38: Pull these from args at the top, it should be in scope

```
{ buildPackages,
, dtc
, ftpmHelperTa
, gitRepos
...
```

pkgs/optee/taDevKit.nix
```diff
@@ -1,5 +1,8 @@
...
+  ftpmHelperTa = null;
+  msTpm20RefTa = null;
```
elliotberman, line 6: I think we only need to set enableFTPM = false and the others are unnecessary overrides.

modules/optee.nix
```diff
@@ -108,43 +167,140 @@ in
...
+          requires l4t r36.x. r35 lacks fTPM source; r38 refactored fTPM
+          compilation to use the CFG_MS_TPM_20_REF flag and is not yet
+          supported here.
+        '';
+      } {
+        assertion = !cfg.ftpm.enable || pkgs.nvidia-jetpack.socType == "t234";
+        message = ''
+          hardware.nvidia-jetpack.firmware.optee.ftpm.enable currently
+          requires a t234 (Orin) SoC. Got: ${pkgs.nvidia-jetpack.socType}
+        '';
```
elliotberman, line 186: Is the assertion about JetPack version sufficient? JetPack 6 only supports Orin, so this doesn't catch anything the above assertion can't.

pkgs/optee/opteeClient.nix
```diff
@@ -21,7 +36,17 @@ stdenv.mkDerivation {
...
+    ]);
```
elliotberman, line 49: I think this is out of date?

elliotberman (changes requested):
Thanks for this contribution! The series needs `nix fmt` for pre-submit.

pkgs/optee/opteeClient.nix
TanelDettenborn: I have a comment regarding this section from a slightly different perspective. It would be great to add a clarification comment here to improve maintainability.

I haven't verified r35, but r36 does not include the sd_notify-support patch - yet. It is included in r38, so I assume r39 has it as well.

modules/optee.nix
TanelDettenborn: I believe `config.hardware.nvidia-jetpack.optee.taLogLevel` cannot be used here, because it has a different default value. If NVIDIA sources/BSP are followed closely, then CFG_TEE_TA_LOG_LEVEL for the fTPM TA needs to be set to [0](https://gitlab.com/nvidia/nv-tegra/tegra/optee-src/nv-optee/-/blob/l4t/l4t-r36.5/optee/samples/ms-tpm-20-ref/Samples/ARM32-FirmwareTPM/optee_ta/fTPM/Makefile?ref_type=heads#L8). However, `config.hardware.nvidia-jetpack.optee.taLogLevel` defaults to [2](https://github.com/anduril/jetpack-nixos/blob/master/modules/optee.nix#L102) (even though NVIDIA uses [1](https://gitlab.com/nvidia/nv-tegra/tegra/optee-src/nv-optee/-/blob/l4t/l4t-r36.5/optee/optee_os/mk/config.mk?ref_type=heads#L82)).

As separate comment then CFG_TA_LOG_LEVEL-does not exist. It should be CFG_TEE_TA_LOG_LEVEL.

modules/optee.nix
```diff
@@ -68,6 +70,63 @@ in
...
+            MS TPM 2.0 reference TA build and CFG_CORE_TPM_EVENT_LOG in
+            OP-TEE OS.
+          '';
+        };
+
+        taLogLevel = mkOption {
+          type = types.int;
+          default = 0;
+          description = ''
+            CFG_TA_LOG_LEVEL passed to the MS TPM 2.0 reference TA build.
```
TanelDettenborn, line 104: `CFG_TA_LOG_LEVEL`  should be `CFG_TEE_TA_LOG_LEVEL`

overlay-with-config.nix
```diff
@@ -76,11 +76,26 @@ final: prev: (
       armTrustedFirmware = prevJetpack.armTrustedFirmware.overrideAttrs {
         inherit (finalJetpack) socType;
       };
+
+      msTpm20RefTa = prevJetpack.msTpm20RefTa.overrideAttrs (prevAttrs: {
+        patches = prevAttrs.patches or [ ] ++ cfg.firmware.optee.patches;
+        makeFlags = prevAttrs.makeFlags or [ ]
+          ++ [ "CFG_TA_LOG_LEVEL=${toString cfg.firmware.optee.ftpm.taLogLevel}" ]
```
TanelDettenborn, line 83: `CFG_TA_LOG_LEVEL` should be `CFG_TEE_TA_LOG_LEVEL`

zacharyweiss:
Superseded by merged #521; should be closed

PR #499, opened 2026-05-08
kernels: r35: Add patch names

elliotberman:
###### Description of changes

Adding names to the patches, since `kernelPatches` expects to have a list of attribute sets with `name` and `patch` defined.

###### Testing

- [x] hydra eval passes


ConnorBaker:
@elliotberman verified with #498 that this solves the eval failures I saw.

ConnorBaker (approved):

elliotberman:
superseded by #498

PR #505, opened 2026-05-21
optee: Apply Overlayed OP-TEE Patches to opteeClient

Hussainity:
###### Description of changes

This PR applies the OP-TEE patches exposed by `hardware.nvidia-jetpack.firmware.optee.patches` to `opteeClient`.
These changes were formerly only applied to `optee-os`. This works because `opteeClient` and `optee-os` share the same underlying `nv-optee` repository.

###### Testing

Verified that changes get applied using a patch targeting OP-TEE Client (used a Jetson AGX board on JP6).

### DIFF
--- a/overlay-with-config.nix
+++ b/overlay-with-config.nix
@@ -82,6 +82,9 @@ final: prev: (
         patches = prevAttrs.patches or [ ] ++ cfg.firmware.optee.patches;
         makeFlags = prevAttrs.makeFlags or [ ] ++ cfg.firmware.optee.extraMakeFlags;
       });
+      opteeClient = prevJetpack.opteeClient.overrideAttrs (prevAttrs: {
+        patches = (prevAttrs.patches or [ ]) ++ cfg.firmware.optee.patches;
+      });
 
       flashInitrd =
         let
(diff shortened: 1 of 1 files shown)


overlay-with-config.nix
```diff
@@ -82,6 +82,9 @@ final: prev: (
         patches = prevAttrs.patches or [ ] ++ cfg.firmware.optee.patches;
         makeFlags = prevAttrs.makeFlags or [ ] ++ cfg.firmware.optee.extraMakeFlags;
       });
+      opteeClient = prevJetpack.opteeClient.overrideAttrs (prevAttrs: {
+        patches = (prevAttrs.patches or [ ]) ++ cfg.firmware.optee.patches;
```
elliotberman: Do we care about the other overrideAttrs we do for optee-os?

Hussainity: Didn't need it so I didn't include it - but I can see an argument for symmetry? Thoughts?

Princemachiavelli: I'd like makeFlags too. I actually started a PR just like this but I wasn't sure if we wanted to force a rebuild of the OPTEE OS and OPTEE client when either of their patches/makeFlags change or if we should have separate options. I'm open to either implementation; I think since we don't change these patches or flags frequently it's fine for them to use the same options.

Princemachiavelli (approved):
Approved but would like `makeFlags` to apply to OPTEE client too.

elliotberman (approved):

PR #509, opened 2026-06-01
Add equivalent of 0003-Lower-priority-of-tegra-se-crypto.patch for JP6

ktrinh-anduril:
###### Description of changes
Add equivalent of 0003-Lower-priority-of-tegra-se-crypto.patch for JP6

###### Testing
Local testing with a netboot install shows without the patch, the installation is significantly slower on LUKS rootfs


elliotberman:
Duplicate of #502, which also lowers prio on JP7

PR #513, opened 2026-06-09
udev: rename tegra-devices rules from 99 to 90 prefix

sm1ll3r:
###### Description of changes

The L4T udev rules were installed as 99-tegra-devices.rules. NixOS places rules from services.udev.extraRules into 99-local.rules, which sorts lexicographically *before* 99-tegra-devices.rules. This means the tegra rules run last and win, making it impossible to easily override them via extraRules.

Rename the file to 90-tegra-devices.rules so it is processed earlier in the udev rule ordering, allowing users to override tegra device rules through the standard services.udev.extraRules mechanism (which lands in 99-local.rules).

###### Testing

Verified rules file was renamed and device is booting without issue on:
- [x] orin-agx-devkit-jp5
- [x] orin-agx-devkit-jp6
- [x] thor-agx-devkit

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -370,7 +370,8 @@ in
 
       services.udev.packages = [
         (pkgs.runCommand "jetson-udev-rules" { } ''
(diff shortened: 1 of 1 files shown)


elliotberman (approved):
LGTM, please test on JetPack 5, 6, and 7.

PR #516, opened 2026-06-11
flash-script: make initrd-flash host-side timeouts configurable

vadika:
Replace the hardcoded initrd-flash host-side timeouts with NixOS options so they can be raised for slow hardware without patching the module:

- `initrdFlashSerialEnumerateTimeoutSeconds` (default 240s) — wait for the booted initrd's USB serial to appear at `/dev/serial/by-id/...`.
- `initrdFlashCompletionTimeoutSeconds` (default 900s) — `expect` watcher timeout for the flash-success/-failure line on the device serial console.

Defaults preserve historical behavior. Raise them for slow QSPI parts or long firmware flashes.

Eval-tested against `orin-agx-devkit` (options resolve to 900/240). `treefmt` formatting check passes clean. Not exercised on real hardware (needs aarch64).

### DIFF
--- a/device-pkgs/initrdflash-script.nix
+++ b/device-pkgs/initrdflash-script.nix
@@ -118,7 +118,7 @@ let
     # --log-strip -> Strip control characters and escape sequences
     # --mute -> Mute tio messages
     spawn -noecho "${lib.getExe tio}" --log-strip --mute "$modem"
-    set timeout 900
+    set timeout ${builtins.toString cfg.flashScriptOverrides.initrdFlashCompletionTimeoutSeconds}
     set failed 0
     expect {
       "Flashing platform firmware successful" {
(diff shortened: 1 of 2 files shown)


elliotberman:
Hi @vadika , I'm trying to understand motivation for the change a bit better. What is a signed SD image? Can the signed artifacts be in the Nix store?

We're tying to not support all the different ways of potentially flashing a device since that gets really quite complicated quickly; instead focusing on the "easy" path and then adding hooks or overriding as requested (example: `hardware.jetpack-nixos.flashScriptOverrides.preFlashCommands`). I'm very open to adding hooks in this form. If this should be a commonly supported feature, then I think it might make sense to have dedicated options -- that's the angle I'm trying to understand with the first couple questions above.

vadika:
> Hi @vadika , I'm trying to understand motivation for the change a bit better. What is a signed SD image? Can the signed artifacts be in the Nix store?
> 
> We're tying to not support all the different ways of potentially flashing a device since that gets really quite complicated quickly; instead focusing on the "easy" path and then adding hooks or overriding as requested (example: `hardware.jetpack-nixos.flashScriptOverrides.preFlashCommands`). I'm very open to adding hooks in this form. If this should be a commonly supported feature, then I think it might make sense to have dedicated options -- that's the angle I'm trying to understand with the first couple questions above.

Sorry for misleading, the second commit was not related to this PR, the scope is just the timeouts tunabled.

elliotberman (approved):
Thanks!

Issue #507, opened 2026-05-27
JetPack 6 / CUDA 12.x compute libraries not available for aarch64 — blocks native Python ML workloads


vonjackets:
Hey all, I've been experimenting with this project in order to attempt some AI benchmarking efforts on the Jetson Orin AGX and put some work into trying to get around the containerization approach that's become popular over the years. It seems like this project has made many leaps forward to get us there, but I ran into some issues and hope to get some feedback here. 

When attempting to run CUDA-accelerated Python packages (specifically `torch`, `ctranslate2`, and `onnxruntime-gpu` with CUDA provider) natively on a JetPack 6 NixOS system, the packages fail to import because the CUDA 12.x compute libraries they depend on at runtime are not present in the Nix store for aarch64.

I believe this is a gap between what the JetPack BSP provides (the CUDA runtime via `nvidia-l4t-cuda`) and what the full CUDA compute stack requires (libcublas, libcudnn, libcufft, libcurand, libcusparse, libcusolver, libnvrtc, libnvJitLink, libcupti, and related libraries).

On Ubuntu JetPack these are installed via NVIDIA's CUDA apt repository (`cuda-toolkit-12-x`) as a separate step from the BSP. On NixOS there is no equivalent — the nixpkgs CUDA redistributables are x86_64 only, and jetpack-nixos does not currently package the aarch64 JetPack 6 CUDA compute library debs.

---

## Environment

- Device: Jetson AGX Orin devkit
- JetPack version: 6.2 (L4T 36.5.0)
- NixOS via jetpack-nixos master
- nixpkgs: nixos-25.11

---

## How I Got Here

I was attempting to package a whisperX speech recognition benchmarking pipeline natively in a Nix devShell targeting JetPack 6, avoiding the container-based approach used on Ubuntu. The dependency chain that exposed this gap:

**whisperX → faster-whisper → ctranslate2 → CUDA compute libs**
**whisperX → torch → CUDA compute libs**

When `import torch` is called, torch's `_load_global_deps()` function attempts to `ctypes.CDLL` its bundled `libtorch_global_deps.so`. This fails because `libcudart.so.12` is not findable. It then falls into a recovery path that searches `sys.path` for NVIDIA wheel-layout directories (e.g. `{site-packages}/nvidia/cuda_runtime/lib/libcudart.so.*`). These don't exist either. The full list of libraries torch attempts to preload:

```
libcublas.so.*[0-9]
libcudnn.so.*[0-9]
[...]

PR #525, opened 2026-07-06
Package CUDA 12.6 compute libs for aarch64 JetPack 6 (#507)

Abhishek21g:
## Summary

- Add `cuda-packages-12-6-extension.nix` to wire Jetson CUDA 12.6 compute debs from `sourceinfo/r36.5-debs.json` into `cudaPackages_12_6` on `linux-aarch64`, using the existing `debWrapBuildRedist` pattern from JetPack 5 / CUDA 11.4
- Register the extension in `overlay.nix` so `libcublas`, `cudnn`, `libcufft`, `cuda_nvrtc`, `libnvjitlink`, `cuda_cupti`, `libcufile`, and related libraries are available in the Nix store for native ML workloads (PyTorch, ctranslate2, etc.)
- Add `examples/jp6-ml-devshell/flake.nix` and README documentation for the JetPack 6 native ML workflow

Closes the packaging gap described in #507: the debs were already in the Jetson manifest but not exposed through `cudaPackages` for CUDA 12.6 on aarch64.

Known follow-ups (not in Jetson deb manifest): `nccl`, `libcusparseLt` — may need SBSA apt repo packaging.

## Test plan

- [ ] `nix eval .#legacyPackages.aarch64-linux.cudaPackages_12_6.libcublas.name` — should **not** return `cuda11.4-libcublas-*`
- [ ] `nix build .#legacyPackages.aarch64-linux.cudaPackages_12_6.libcublas`
- [ ] `cd examples/jp6-ml-devshell && nix run .#smoke-eval` on aarch64
- [ ] On Orin AGX (JP 6.2 / L4T 36.5.0): `nix develop` → `python -c "import torch; print(torch.cuda.is_available())"` with torch wheel + `LD_LIBRARY_PATH` or wheel-layout symlinks
- [ ] Confirm Hydra `orin.nvidia-jetpack6` package set still evaluates

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,6 +325,54 @@ To make `jetpack-nixos`'s CUDA package set the default, provide Nixpkgs with thi
 final: _: { inherit (final.nvidia-jetpack) cudaPackages; }
 ```
 
+### JetPack 6 native ML (CUDA 12.6 compute libraries)
+
+JetPack 6 systems should use `cudaPackages_12_6` (not the default `cudaPackages_11_4`). On `aarch64-linux`, the overlay replaces upstream CUDA redistributables with Jetson debs from `sourceinfo/r36.5-debs.json` — the same `debWrapBuildRedist` pattern used for CUDA 11.4 on JetPack 5.
+
+This wires `libcublas`, `cudnn`, `libcufft`, `cuda_nvrtc`, `libnvjitlink`, `cuda_cupti`, and related libraries into the Nix store so native Python ML packages (e.g. PyTorch) can load CUDA dependencies without containers.
+
+**Verify on an Orin (JetPack 6):**
+
+```shell
+# After enabling the jetpack-nixos overlay and cudaPackages_12_6:
+nix eval .#legacyPackages.aarch64-linux.cudaPackages_12_6.libcublas.name
+# Should NOT be cuda11.4-libcublas-*
+
+nix build .#legacyPackages.aarch64-linux.cudaPackages_12_6.libcublas
+```
+
+**devShell sketch** (see `examples/jp6-ml-devshell/flake.nix`):
+
+```nix
+{
+  inputs.jetpack-nixos.url = "github:anduril/jetpack-nixos";
+  outputs = { jetpack-nixos, nixpkgs, ... }: {
+    devShells.aarch64-linux.default = let
+      pkgs = import nixpkgs {
+        system = "aarch64-linux";
+        config = {
+          allowUnfree = true;
+          cudaSupport = true;
+          cudaCapabilities = [ "8.7" ];
+        };
+        overlays = [
+          jetpack-nixos.overlays.default
+          (final: _: { cudaPackages = final.cudaPackages_12_6; })
+        ];
+      };
+    in pkgs.mkShell {
+      packages = with pkgs.cudaPackages; [
+        libcublas cudnn libcufft cuda_cudart cuda_nvrtc libnvjitlink
+      ];
+    };
+  };
+}
+```
+
+PyTorch wheels may still expect libraries under `site-packages/nvidia/*/lib/`; symlink or set `LD_LIBRARY_PATH` to the `cudaPackages` outputs until wheel layout helpers land.
+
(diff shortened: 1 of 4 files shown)


Abhishek21g:
**Validation status (author):** No Orin AGX hardware in my dev environment yet — aarch64 eval/build smoke tests in the PR test plan are pending on-device confirmation.

**Ready for reviewer with hardware:**
```bash
nix eval .#legacyPackages.aarch64-linux.cudaPackages_12_6.libcublas.name
nix build .#legacyPackages.aarch64-linux.cudaPackages_12_6.libcublas
cd examples/jp6-ml-devshell && nix run .#smoke-eval
nix develop  # then torch import smoke per README
```

**Separate product track** (not a substitute for the above): [Jetson ML Workbench](https://enaguthi.com/jetson-ml-workbench/site/) models #482-style closure drift + SLA gating for perception deploys — independent of this packaging PR.

elliotberman:
Hi @Abhishek21g , thanks for raising this!

smoke-eval and devshell both builds for me without the additional `cuda-packages-12-6-extension.nix`. I was also able to run the smoke-eval. Should this have failed without the rest of the PR?

Abhishek21g:
Thanks @elliotberman — fair question, and you're right to call this out.

**Short answer:** No — `smoke-eval` as written should *not* necessarily fail without `cuda-packages-12-6-extension.nix`. It only checks that `cudaPackages_12_6.libcublas` evaluates/builds and that `libcublas.so*` exists in the store. On current `nixos-25.11`, nixpkgs already ships CUDA 12.6 redistributables for `linux-aarch64`, so the devShell and smoke script can pass using upstream redist-backed derivations alone.

**What the extension is actually for:** mirroring the existing `cuda-packages-11-4-extension.nix` pattern — on `aarch64-linux` + CUDA 12.6, replace nixpkgs redistributables with Jetson debs from `sourceinfo/r36.5-debs.json` via `debWrapBuildRedist`, so the Nix closure matches the L4T/JetPack BSP stack rather than generic SBSA redists. That's a closure-alignment change, not "make libcublas exist at all."

**Where I overclaimed:** #507 reads like compute libs are missing entirely; on today's nixpkgs that's likely overstated for eval/build. The remaining gap is probably narrower — e.g. wrong default `cudaPackages` (11.4 stubs), runtime torch/ORT wiring, or Jetson-vs-redist ABI drift — and the original reporter hasn't posted the minimal repro you asked for on #507 yet.

**Discriminator I'd like to compare on your Orin build (with vs without the extension line in `overlay.nix`):**
```bash
nix eval .#legacyPackages.aarch64-linux.cudaPackages_12_6.libcublas.name
nix eval .#legacyPackages.aarch64-linux.cudaPackages_12_6.libcublas.src --raw 2>/dev/null || true
```
With the extension, I'd expect deb/Jetson-sourced builds; without it, redist tarballs from `developer.download.nvidia.com/compute/cuda/redist/...`.

Happy to revise the PR accordingly:
1. Tighten `smoke-eval` to assert deb-backed / Jetson-sourced derivations (not just `.so` presence)
2. Reframe README + test plan around Jetson closure parity vs nixpkgs aarch64 redists
3. Narrow or drop anything redundant if your eval shows no meaningful delta

Could you share the `libcublas.name` (and ideally whether torch import works) on your side without the extension? That'll tell us whether this should merge as-is, shrink, or close.

elliotberman:
> the Nix closure matches the L4T/JetPack BSP stack rather than generic SBSA redists.

Nope, we're getting the linux-aarch64 redists not the SBSA ones.

> With the extension, I'd expect deb/Jetson-sourced builds; without it, redist tarballs from `developer.download.nvidia.com/compute/cuda/redist/...`.

Yep, this is expected.

Abhishek21g:
got it — thanks for clarifying. sounds like aarch64 redists are already the right path on current nixpkgs and smoke-eval doesn't need the extension. i'll hold off on more changes unless you think jetson deb closure is still worth pursuing for #507, or happy to close if the issue is stale without vonjackets' repro.

Abhishek21g:
closing this for now — appreciate the review.

PR #537, opened 2026-08-08
Support fusing Thor family using FSKP

ssloboda-anduril:
###### Description of changes

Adds baseline support for secure boot on thor family.

###### Testing

Built and run on thor-agx-devkit.


pkgs/flash-tools/default.nix
```diff
@@ -149,6 +157,7 @@ let
       openssl
       sbsigntool # In l4t_uefi_sign_image.sh, which is needed by RCM flashing
       xmlstarlet # Needed in JetPack 7
+      libxml2 # Needed in JetPack 7
```
elliotberman: What is needed from libxml2? Is it `xmllint`?

pkgs/flash-tools/default.nix
```diff
@@ -126,6 +130,10 @@ let
       runHook postInstall
     '';
 
+    # TODO I think these should be pulled in via `runtimeInputs = flashtools.passthru.flashDeps`
+    # rather than hacking them into the text of the flash-script style scripts. Hacking in via
+    # `PATH` also isn't correct for libraries.
```
elliotberman: Agreed, we can leave this as TODO for later. Although I don't think we're "hacking in via PATH" for libraries currently

overlay.nix
```diff
@@ -55,19 +55,47 @@ in
     final
     prev;
 
-  nvidia-jetpack7 = import ./mk-overlay.nix
-    {
-      jetpackMajorMinorPatchVersion = "7.2";
+  # "JetPack 7 and 7.1 is r38.x, and uhh JetPack 7.2 is r39. Isn't it great?"
+  # - Anonymous
```
elliotberman: Drop

overlay.nix
```diff
@@ -55,19 +55,47 @@ in
...
+
+        bspHash = "sha256-FiZibNgn3g41C4gCAzudplPGmyKQrM7bnl0B9JYH4Jk=";
+        bspPrePatch = ''
+          cp --no-preserve=all -r ${hostOverlayFskpTools}/Linux_for_Tegra/. ./
+          find ${hostOverlayFskpTools}/Linux_for_Tegra/. -type f -perm -u+x -printf '%P\0' \
+            | xargs -0 -I{} chmod +x "./{}"
+
+          # Remove NVIDIA's bundled pyusb so we use nix's version. See `fskp_helper_t264.py`.
+          find l4t/tools/flashtools -type d -name pyusb -path '*/external/pyusb' -exec rm -rf {} +
+        '';
```
elliotberman, line 89: Does applying FSKP overlay prevent Orin from using odmfuse.sh? If yes, does it make sense to apply the host overlay only when `fskp.enable == true`?

modules/flash-script.nix
```diff
@@ -180,6 +180,17 @@ in
...
+          pkcList = mkOption {
+            type = types.nullOr types.package;
+            default = null;
+            description = ''
+              Package containing NVIDIA's PKC list xml file containing the PKCs used to sign
+              firmware and in the board's combination PKC hash. Setting this value automatically
+              populates pkcFile. Use the `mkPkcList` helper function to create a pkc list file with
+              the quirks required for the nix build.
+            '';
+          };
```
elliotberman: Drop this option, users should set `pkcFile` directly

```nix
let
   pkcList = mkPkcList { ... };
in
{
    hardware.nvidia-jetpack.firmware.secureBoot.pkcFile = "${pkcList}/pkc_keylist.xml";
}
```

pkgs/flash-tools/default.nix
```diff
@@ -39,6 +39,10 @@ let
     pip
     python-pkcs11
     cryptography
+    # TODO do we really want to pull in libusb for all flash-script style scripts?
+    # Needed in JetPack 7 to replace NVIDIA's circa 2016 bundled pyusb/libusb1
```
elliotberman: Yes. Please drop the TODO line but the second line is good docs to keep.

pkgs/flash-tools/default.nix
ssloboda-anduril: I would have needed to add a `LD_LIBRARY_PATH` for libusb iirc, so wording is probably just wrong

overlay.nix
ssloboda-anduril: I think you might be able to use odmfuse.sh for a couple of the fuses on thor (not any of the ones that contains secrets) but I did not test it since it didn't seem useful. FSKP does not prevent usage of odmfuse.sh on orin. The overlay puts the fskp tools under tools/flashtools/ which is otherwise not existent in the bsp. I don't see the value in making it conditional - just extra complexity

modules/flash-script.nix
ssloboda-anduril: I think `pkcList` is used to generate the combo hash. So i guess we'd be relying on not evaling that and just using `pkcFile` instead in that derivation. I think it is useful to have direct access to the pkcList and easily be able to check if we have a xml file or a single keyfile.

Thoughts?

ssloboda-anduril: ah nevermind, i was mistaken. removing

pkgs/flash-tools/default.nix
ssloboda-anduril: yes

ssloboda-anduril: i see we have a different way of handling for jetpack5. Which is prefered

ssloboda-anduril:
retested on device, i think i addressed the comments. lmk if more is needed

elliotberman (approved):

PR #538, opened 2026-08-19
cudaPackages: tensorrt: Clear executable permission bit from stack in libraries

elliotberman:
###### Description of changes

TensorRT's shared objects (notably libnvinfer_builder_resource.so) request an executable stack (PT_GNU_STACK = RWE). glibc >= 2.41 makes dlopen() of such objects a hard error instead of silently granting an executable stack, and TensorRT lazily dlopen()s the builder resource at engine-build time, so this surfaces as an opaque runtime failure:

Error Code 6: Internal Error
(Unable to load library: libnvinfer_builder_resource.so.8.5.2)

Clear the exec-stack bit from the libs so they load on newer glibc.

###### Testing

Observe exec stack bit cleared on:
- [x] `readelf -l result-lib/lib/libnvinfer_builder_resource.so.10.7.0` (cuda 12)
- [x] `result-lib/lib/libnvinfer_builder_resource.so.8.5.2` (cuda 11)

### DIFF
--- a/cuda-packages-11-4-extension.nix
+++ b/cuda-packages-11-4-extension.nix
@@ -202,7 +202,7 @@ let
             ];
 
             postFixup =
-              prevAttrs.prevAttrs or ""
+              prevAttrs.postFixup or ""
               + ''
                 nixLog "patchelf-ing ''${!outputLib:?}/lib/libnvinfer.so with runtime dependencies"
                 patchelf \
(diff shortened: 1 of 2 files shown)


danielfullmer:
Do you happen to know whether tensorrt actually depends on having an executable stack in some code paths? I see others on the nvidia forum doing this same workaround, but have you seen if nvidia has ever confirmed this is OK?

elliotberman:
> I see others on the nvidia forum doing this same workaround, but have you seen if nvidia has ever confirmed this is OK?

Yes, an NVIDIA employee seems to have confirmed this is the right solution: https://forums.developer.nvidia.com/t/tensorrt-10-8-0-43-and-older-incompatible-with-new-glibc-2-41-r2-g0a7c7a3e283a/322769/6

PR #539, opened 2026-08-20
uefi: add firmware.uefi.edk2ExtraPackages

rlfishback:
###### Description of changes

Add `hardware.nvidia-jetpack.firmware.uefi.edk2ExtraPackages`, allowing boards to add out-of-tree EDK2 package trees to the UEFI build workspace.
Each package is unpacked into the workspace and added to EDK2's `PACKAGES_PATH` through Stuart's `--insert-packages-path`. This lets board-specific DSC/FDF patches reference packages such as `IntelUndiPkg` without patching an existing EDK2 source tree.

The option defaults to an empty list, so existing UEFI builds are unchanged. It is restricted to L4T r36+, since the r35 firmware builder does not support the Stuart-based package-path flow. Evaluation assertions reject duplicate package names and collisions with built-in EDK2 source trees.

###### Testing

- [x] `nix flake check --no-build`
- [x] Build JP7 UEFI firmware with a non-empty `edk2ExtraPackages`
- [x] Confirm an external EDK2 driver package is compiled into the resulting firmware build

### DIFF
--- a/modules/flash-script.nix
+++ b/modules/flash-script.nix
@@ -66,6 +66,15 @@ in
             default = [ ];
           };
 
+          edk2ExtraPackages = mkOption {
+            type = types.listOf types.package;
+            description = ''
+              Extra edk2 package trees for the UEFI build. Each derivation is
+              unpacked under its name and prepended to PACKAGES_PATH. Requires L4T r36 or newer.
+            '';
+            default = [ ];
+          };
+
           secureBoot = {
             enrollDefaultKeys = lib.mkEnableOption "enroll default UEFI keys";
             defaultPkEslFile = mkOption {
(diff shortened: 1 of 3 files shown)


pkgs/uefi-firmware/stuart.nix
```diff
@@ -167,7 +186,7 @@ lib.extendMkDerivation {
 
       buildPhase = ''
         stuart_setup -c "edk2-nvidia/Platform/NVIDIA/${platformBuild}/PlatformBuild.py"
-        stuart_build -c "edk2-nvidia/Platform/NVIDIA/${platformBuild}/PlatformBuild.py" ${stuartExtraArgs} --target ${buildTarget}
+        stuart_build -c "edk2-nvidia/Platform/NVIDIA/${platformBuild}/PlatformBuild.py"${extraPackagesPathFlags} ${stuartExtraArgs} --target ${buildTarget}
```
elliotberman:
```suggestion
        stuart_build -c "edk2-nvidia/Platform/NVIDIA/${platformBuild}/PlatformBuild.py" ${extraPackagesPathFlags} ${stuartExtraArgs} --target ${buildTarget}
```

modules/flash-script.nix
```diff
@@ -463,6 +472,10 @@ in
         assertion = (cfg.firmware.fskp.enable && (cfg.firmware.fskp.fuseBlob ? insecureClearText) -> cfg.firmware.fskp.fuseBlob.insecureClearText);
         message = "Do not set fskp.fuseBlob.insecureClearText in order to use an encrypted blob";
       }
+      {
+        assertion = cfg.firmware.uefi.edk2ExtraPackages == [ ] || pkgs.nvidia-jetpack.l4tAtLeast "36";
```
elliotberman: It's logically equivalent, but implication is more expressive in my opinion. What do you think?

```suggestion
        assertion = cfg.firmware.uefi.edk2ExtraPackages != [ ] -> pkgs.nvidia-jetpack.l4tAtLeast "36";
```

elliotberman (approved):

PR #542, opened 2026-08-24
Bump JetPack 7 from JP7.2 / L4T 39.2.0 to JP7.2.1 / L4T 39.2.1

rlfishback:
This updates the regenerated JP7 source metadata and package versions for the new NVIDIA release. It also updates JP7 UEFI edk2 source fetches to the matching `r39.2.1` tags. Only `edk2-nvidia` changed source content, so it is the only updated fixed-output hash.

###### Testing

  - [x] Orin Nano devkit
  - [x] Orin NX devkit
  - [x] Orin AGX devkit
  - [x] Orin AGX Industrial devkit
  - [x] Thor AGX devkit


pkgs/containers/l4t-csv.nix
```diff
@@ -36,4 +36,9 @@ runCommand "l4t-csv"
   mkdir -p $out
   dpkg --fsys-tarfile ${debs.${repo}.nvidia-l4t-init.src} | tar -x ./etc/nvidia-container-runtime/host-files-for-container.d
   cp etc/nvidia-container-runtime/host-files-for-container.d/* $out/
+
+  # Include the libcuda payload required by CDI.
+  if [[ ${if l4tAtLeast "39" then "1" else "0"} == 1 ]]; then
+    echo 'lib, /usr/lib/aarch64-linux-gnu/nvidia/libcuda.so.1.1' >> $out/drivers.csv
+  fi
```
elliotberman: Use lib.optionalString

pkgs/containers/deps.nix
```diff
@@ -4,26 +4,38 @@
...
+          '')
+          debFiles)))
+        (lib.importJSON ./r${lib.versions.major l4tMajorMinorPatchVersion}-l4t.json)))
+    /* Temporarily disabled to test JP7.2.1 without the libcuda CDI workaround.
+      + lib.optionalString (l4tAtLeast "39") ''
+      # Provide libcuda at the location used by the CDI CSV.
+      mkdir -p $out/usr/lib/aarch64-linux-gnu/nvidia
+      ln -s ${l4t-cuda}/lib/libcuda.so.1.1 $out/usr/lib/aarch64-linux-gnu/nvidia/libcuda.so.1.1
+      ln -s libcuda.so.1.1 $out/usr/lib/aarch64-linux-gnu/nvidia/libcuda.so.1
+      ln -s libcuda.so.1 $out/usr/lib/aarch64-linux-gnu/nvidia/libcuda.so
```
elliotberman, line 38: Ah. NVIDIA is also generating the CSV on the fly based on whether using the openrm/nvgpu stack. We take care of this automatically inside `l4t-cuda`, so we don't need to mirror the logic here, we just have to apply the effect. What's here is correct for libcuda.so.

I found some additional libraries we might need to also be incorporating (exactly the libraries being deleted from the r39-l4t.json).

See etc/systemd/nv-load-gpu-libs.sh in the nvidia-l4t-init deb.

Please wire in these additional libraries to the CSV and also add comment explaining to check this script for the host libraries that should also be exposed to the container.

```sh
        libcuda_real="$(nv_load_gpu_libs_resolve_regular_file \
                "${lib_dir}/libcuda.so.1.1")" || {
                echo "ERROR: nv-load-gpu-libs: missing libcuda.so.1.1 under ${lib_dir}" >&2
                return 1
        }
        inst_real="$(nv_load_gpu_libs_resolve_regular_file \
                "${lib_dir}/libcuda_instrumentation.so")" || {
                echo "ERROR: nv-load-gpu-libs: missing libcuda_instrumentation.so under ${lib_dir}" >&2
                return 1
        }
        if [ "${suffix}" = "openrm" ] && [ -e "${lib_dir}/libnvcuvid.so" ]; then
                nvcuvid_real="$(nv_load_gpu_libs_resolve_regular_file \
                        "${lib_dir}/libnvcuvid.so")" || {
                        echo "ERROR: nv-load-gpu-libs: invalid libnvcuvid.so under ${lib_dir}" >&2
                        return 1
                }
        fi

```

elliotberman (changes requested):

pkgs/containers/l4t-csv.nix
```diff
@@ -36,4 +36,13 @@ runCommand "l4t-csv"
   mkdir -p $out
   dpkg --fsys-tarfile ${debs.${repo}.nvidia-l4t-init.src} | tar -x ./etc/nvidia-container-runtime/host-files-for-container.d
   cp etc/nvidia-container-runtime/host-files-for-container.d/* $out/
+
+  ${lib.optionalString (l4tAtLeast "39") ''
+  # NVIDIA generates these entries at boot in
+  # /etc/systemd/nv-load-gpu-libs.sh. Check that script for host libraries
+  # that must also be exposed to containers.
```
elliotberman, line 44: Where does that script live? in `debs.som.nvidia-l4t-init`

pkgs/l4t/l4t-nvpmodel.nix
```diff
@@ -1,8 +1,18 @@
...
   buildInputs = [ l4t-core ];
+
+  postFixup = ''
+    # The kernel no longer provides the legacy EMC/bwmgr controls.
+    if [[ ${if l4tAtLeast "39" then "1" else "0"} == 1 ]]; then
+      for conf in $out/etc/nvpmodel/*.conf; do
+        sed -i -e '/bwmgr/d' -e '/EMC/d' "$conf"
+      done
+    fi
+  '';
```
elliotberman: I'd like to seem some evidence that L4T is doing similar thing. I'd guess that the EMC/bwmgr controls are still expected to exist but we might be not loading some kernel module or setting some kernel param.

elliotberman (changes requested):

pkgs/containers/l4t-csv.nix
rlfishback: It is installed from debs.som.nvidia-l4t-init, at /etc/systemd/nv-load-gpu-libs.sh

pkgs/l4t/l4t-nvpmodel.nix
rlfishback: L4T 39.2.1 still expects EMC/bwmgr controls. We were including the module package but not explicitly loading the module. I removed the workaround that stripped NVIDIA’s EMC entries from nvpmodel and load tegra-bpmp-bwmg on JP7 instead. I’ll confirm this on HITL.

rlfishback: tegra-bpmp-bwmg loads automatically, don't need to explicitly add it.

elliotberman (approved):

PR #543, opened 2026-08-27
Fix regression in thor fuse & flash scripts

ssloboda-anduril:
###### Description of changes

* Fixes incorrect path and repo layout assumptions made by the bsp for custom board config files in the fskpFuseScript
* Enables fskp by default on thor family devices (it is the preferred/supported fuse method)
* Implements sign/encrypt now, flash later functionality in flashScript for thor family devices

###### Testing

* Tested flashScript on thor-agx-devkit and confirmed that encryption/signing does not happen at flash-time for single firmware variant configs
* Tested fskpFuseScript dry run on thor-agx-devkit with a custom board conf and confirmed that it ran successfully due to the proper board config being available

### DIFF
--- a/pkgs/r39-bsp-flash-script-no-flash.patch
+++ b/pkgs/r39-bsp-flash-script-no-flash.patch
@@ -1,0 +1,297 @@
+From b706d9aa4b711c426acaf43088085924aa6fbfe7 Mon Sep 17 00:00:00 2001
+From: Steven Sloboda <ssloboda@anduril.com>
+Date: Wed, 19 Aug 2026 11:29:07 -0700
+Subject: [PATCH] Fix thor sign/encrypt now, flash later.
+
+Jetpack 7 flash.sh omits support for using prebuilt firmware artifacts
+in the flash script for thor. Restore this functionality. Eventually, we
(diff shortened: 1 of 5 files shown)


pkgs/r39-bsp-flash-script-no-flash.patch
```diff
@@ -0,0 +1,259 @@
...
+ 	cp -f "${tmp}" "./${localcfgfile}.tmp"
+ 
+@@ -475,6 +478,45 @@ odmsign_ext_flash ()
+ 		fi;
+ 		SOSARGS="--applet rcm_2_${tmp2}.rcm --applet_softfuse rcm_1_${tmp2}.rcm ";
+ 	fi;
++
++	if [ "${CHIPID}" = "0x26" ] || [ "${CHIPID}" = "0x26 8" ]; then
++		if [ "${rcm_tbcfile}" = "" ]; then
++			get_value_from_PT_table "cpu-bootloader" "filename" ${securecfgfile} flashername
```
elliotberman:
```suggestion
+			get_value_from_PT_table "A_cpu-bootloader" "filename" ${securecfgfile} flashername
```

elliotberman (approved):

PR #546, opened 2026-09-01
uefi-firmware: Allow specifying custom DebugPrintErrorLevel

elliotberman:
###### Description of changes

Useful to enable/disable certain prints when debugging issues with the UEFI firmware.

###### Testing

- [x] Set custom error level and observe additional prints


modules/flash-script.nix
```diff
@@ -50,6 +50,12 @@ in
             default = cfg.firmware.uefi.debugMode;
           };
 
+          printErrorLevel = mkOption {
+            type = types.nullOr types.str;
+            default = if cfg.firmware.uefi.errorLevelInfo then "0x8000004F" else null;
```
danielfullmer, line 55: Would be better if this was a human-meaningful string (e.g. info/debug/etc), not just a hex value

danielfullmer: Then this would replace errorLevelInfo entirely.

elliotberman: > Would be better if this was a human-meaningful string (e.g. info/debug/etc), not just a hex value

I don't think so; it's not an enum so there's not a direct string -> number mapping.

danielfullmer: PcdDebugPrintErrorLevel is a bitmask field with values described here: https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Library/DebugLib.h#L36

So maybe a bunch of `.enable` options?

elliotberman: > So maybe a bunch of .enable options?

Understanding a bitmask is probably more ergonomic for developers poking in this area. If this was a high-touch area, I'd probably agree about creating the options.

Princemachiavelli (approved):

PR #547, opened 2026-09-01
patchfv: fix uefi-firmware-parser pname and version

NickCao:
###### Description of changes

To satisfy pythonMetadataCheckHook,
See also https://github.com/NixOS/nixpkgs/commit/d0008c43

Also update edk2-pytool-extensions to 0.29.8 to drop the use of pkg_resources

###### Testing

Built `config.system.build.uefiCapsuleUpdate` as part of a nixos configuration using nixpkgs unstable.


elliotberman:
Ah, I think 1.16 is available in the [26.05 nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-26.05/pkgs/development/python-modules/uefi-firmware-parser/default.nix#L12). We should be able to `nix flake update nixpkgs` and remove the downstream re-packaging.

NickCao:
> Ah, I think 1.16 is available in the [26.05 nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-26.05/pkgs/development/python-modules/uefi-firmware-parser/default.nix#L12). We should be able to `nix flake update nixpkgs` and remove the downstream re-packaging.

Sounds good, I'll try.

NickCao:
> Ah, I think 1.16 is available in the [26.05 nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-26.05/pkgs/development/python-modules/uefi-firmware-parser/default.nix#L12). We should be able to `nix flake update nixpkgs` and remove the downstream re-packaging.

It's also available in the currently pinned version, doesn't even need an update.

elliotberman (approved):

PR #565, opened 2026-09-22
graphics: Fix gpu_pg_mask_param lookup for Thor

burk3:
modprobe.d files are not shell scripts, so the command substitution in the options line was written out verbatim and split into bogus module parameters. gpu_pg_mask_param kept the module's built-in default rather than the value nvpmodel writes to /opt/nvidia/l4t-gpusetup/gpu_pg_mask.

Do the lookup from an install directive instead, which modprobe does run through a shell.

###### Description of changes

replace `options` line (no shell substitution) with an `install` line (yes shell substitution)

###### Testing
before:
```
[root@thor-jp7:~]# dmesg | grep gpu.static
[  105.022500] nv_gpu_static_pg: unknown parameter '$(grep' ignored
[  105.022508] nv_gpu_static_pg: unknown parameter '-E' ignored
[  105.022510] nv_gpu_static_pg: unknown parameter ''^gpu_pg_mask_param' ignored
[  105.022512] nv_gpu_static_pg: unknown parameter '/opt/nvidia/l4t-gpusetup/gpu_pg_mask' ignored
[  105.022513] nv_gpu_static_pg: unknown parameter '||' ignored
[  105.022514] nv_gpu_static_pg: unknown parameter 'echo' ignored
[  105.022514] nv_gpu_static_pg: unknown parameter '512' ignored
[  105.022515] nv_gpu_static_pg: unknown parameter '2>/dev/null)' ignored
```

after: installed with this change on a thor:
```
[root@thor-jp7:~]# dmesg | grep gpu.static
[   13.163482] nv-gpu-static-pg nv-gpu-static-pg: Sending gpu_pg_mask via BPMP MRQ, value = 512.

[root@thor-jp7:~]# lsmod | grep gpu.static
nv_gpu_static_pg       16384  0
```


sm1ll3r (approved):
LGTM

